PR #166, opened 2024-05-10
Refactor/use modular account libs

Zer0dot:
## Motivation
This PR adds the [modular-account-libs](https://github.com/erc6900/modular-account-libs/tree/v0.7.0) node dependency to the project, and shifts common imports to instead import from the newly added dependency, removing redundant files in the process.

This is part of an effort to consolidate dependencies into a shared source, reducing fragmentation across the modular account ecosystem.

## Solution
Removes files and tests for libraries that are now imported from the dependency. These files were diff'd, and are functionally identical to the files already present in `modular-account-libs`, apart from the `FunctionReferenceLib` which had slightly more functionality included in the `modular-account` version. 

This functionality was extracted to a new, simple library `FunctionReferenceHelpers`, but I wouldn't be averse to moving them over to another file if needed. (They're only used once, so we could consider inlining them as well)

### TODO
- [x] Group imports by source (optional, makes import blocks more readable, i.e. Put all modular-account-libs imports in a sequence.)

- [ ] Move the `LLSLRepro.t.sol` file to the `modular-account-libs` repo (optional, but it's the only file which imports tests, and is limited in scope to files contained in the `modular-account-libs` repo)


adamegyed (approved):
Looks good! Thanks for handling some of the more tedious parts of migration.

Long-term, I think we should also migrate our other git submodule dependencies (`forge-std`, `light-account`, the OZ contracts) over to NPM packages as well, but we can tackle that in a later PR.

PR #173, opened 2024-09-05
feat: add code size and test coverage reporting to ci

adamegyed:
## Motivation

Continuous monitoring of test coverage and contract sizes is useful.

## Solution

Add an auto-generated report to each PR against `develop` that shows the code size diff, and test coverage.


.github/workflows/report.yml
```diff
@@ -0,0 +1,112 @@
+name: PR report
```
jaypaik: Super nit: 
```suggestion
name: PR Report
```
Screenshot 2024-09-05 at 6 14 59 PM


To make things consistent with our existing workflows

.github/workflows/report.yml
```diff
@@ -0,0 +1,112 @@
...
+
+on:
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  comment-pr-report:
+    runs-on: ubuntu-latest
```
jaypaik, line 11: Nit: add `name` key and format like the other jobs.

fangting-alchemy (approved):

jaypaik (approved):

.github/workflows/report.yml
adamegyed: Updated! I'm now seeing how these name fields are being used, so I also updated the name in `test.yml` to just be "Test", rather than "Modular Account Test CI", seeing how it is displayed here.

PR #175, opened 2024-09-06
feat: port account tests

adamegyed:
## Motivation

We want to port and update the existing test suite of the RI.

## Solution

Port over the remaining tests in `test/account`. These tests could benefit from a reorganization in the future, as right now they're mostly based on the order in which features were added, rather than areas of account code.

Because the RI uses a test-only factory fixture, but MA uses only the AccountFactory contract, some of the behavior around running SemiModularAccount tests needed to be reorganized.

Also update the test imports.


Zer0dot:
I see the question re: what do we do about SMA coverage remains-- I think it's fine to ignore it for now, and run it manually. 

LGTM! Everything passes. Anything you want us to take a particularly deep look at?

adamegyed:
> SMA coverage

Yeah, I don't think we should implement a "coverage merging" tool across the two test versions, eventually we should find a way to run all test types at once. For now, leaving it as-is.

> Anything you want us to take a particularly deep look at?

Not particularly, just wanted a general look over it!

Zer0dot (approved):
Ty for sorting through the tests we need to port. Lgtm!

PR #176, opened 2024-09-13
feat: gas benchmark setup

adamegyed:
## Motivation

As we optimize MAv2, we want accurate gas benchmarking that can be automatically calculated with low effort, and checked in to VCS. This is following the general plan outlined [here](https://www.notion.so/alchemotion/MAv2-Automation-and-Benchmarking-7d5e67a301b54dc6bf793b5951cf0718#21e5990ce2ab43e5b47ad739600d6105).

## Solution

Install `forge-gas-snapshot`.

Create a new foundry profile called `gas`, with a separate set of tests used to generate the gas snapshots. Included in this PR are ports of the "Runtime: Account Creation" tests from [aa-benchmarks](https://github.com/alchemyplatform/aa-benchmarks), which are both very close the previously observed values in hardhat:
- SimpleAccount creation in foundry: 174143 gas
- SimpleAccount creation in hardhat: 174219 gas
- LightAccount v2 creation in foundry: 143836
- LightAccount v2 creation in hardhat: 143842

I'm not sure how to account for the small remaining gas discrepancies, given the these setups use the exact same bytecode. But the error is small enough that I think we can proceed here.

Also, I'm not sure if we should keep these test cases in this repo long-term, or eventually migrate them somewhere else. I'm leaving them here for now, while we work to create the test cases for MAv2.

Unfortunately, the `forge-gas-snapshot` dependency expects `forge-std` to be remapped differently from the rest of the repo. In theory there should be a way to use foundry's "remappings context" ([more info](https://github.com/foundry-rs/foundry/issues/1855)) to specify this, but I wasn't able to get it to work, so I created separate remappings within `foundry.toml` for the two different profiles.

Also adds a check in CI for having gas snapshots up-to-date, and moves the linting check out to a different workflow file. Both of these still use the shared CI setup action.

Now that there's a fourth linter config, this also moves the solhint config files and the slither config file into a new folder called `config/`, to de-clutter the repo root.

Also adds common actions as scripts in `package.json`:
- `pnpm gas` runs the gas tests and updates values
- `pnpm gas:check` checks that the gas tests have their updated values
- `pnpm fmt` to run `forge fmt` for all folders, and a checker `pnpm fmt:check`
- Updated `pnpm lint` to include the new `gas/` folder.

Delete an accidentally-committed `yarn.lock`.


fangting-alchemy (approved):

gas/lightaccount-v2/LightAccount.gas.t.sol
```diff
@@ -0,0 +1,39 @@
...
+
+import {ILightAccountFactory} from "./ILightAccountFactory.sol";
+import {GasSnapshot} from "forge-gas-snapshot/GasSnapshot.sol";
+import {VmSafe} from "forge-std/src/Vm.sol";
+import {console} from "forge-std/src/console.sol";
+
+contract LightAccountGasTest is GasSnapshot {
+    address internal constant _LIGHT_ACCOUNT_FACTORY = 0x0000000000400CdFef5E2714E63d8040b700BC24;
+
+    bytes internal constant _LIGHT_ACCOUNT_FACTORY_BYTECODE =
```
Zer0dot, line 12: So LightAccount and SimpleAccount are just here for benchmarking right-- so we're not expecting them to get any changes, as the bytecode is frozen completely, yeah?

adamegyed: Yeah, these are frozen at specific version tags. I think we might be OK with removing these in the future, I just added them for now to verify that the gas benchmarking setup works.

PR #177, opened 2024-09-13
feat: Deferred Validation Installation

Zer0dot:
## Motivation

We want to support flows like deferring session key installs until the first action that requires them. 

## Solution

During UO validation, use a specific (previously unused) bit in the signature field to indicate whether or not a deferred validation is necessary. If so, decode part of the signature field into the data needed to validate and install the inner validation. Note that the standard validation packing scheme (for the first 24 bytes of the signature) is, in this case, the validation used to validate the installation.

Post-installation, validate the user operation with the newly installed validation.


Zer0dot:
By the way, the use of the `DeferredValidationInstallData` struct internally literally just solves stack too deep :)

src/account/ModularAccount.sol
```diff
@@ -296,6 +325,10 @@ contract ModularAccount is
         return getAccountStorage().supportedIfaces[interfaceId] > 0;
     }
 
+    function getDomainSeparator() external view returns (bytes32) {
+        return _domainSeparator();
+    }
```
adamegyed: Is there a use for making this split across internal / external copies of the function, or could we make it `public` instead?

src/account/ModularAccount.sol
```diff
@@ -352,32 +366,139 @@ contract ModularAccount is
...
         }
 
-        // Revert if the provided `authorization` less than 21 bytes long, rather than right-padding.
-        ModuleEntity userOpValidationFunction = ModuleEntity.wrap(bytes24(userOp.signature[:24]));
-        bool isGlobalValidation = uint8(userOp.signature[24]) == 1;
+        // Revert if the provided `authorization` less than 24 bytes long, rather than right-padding.
+        ModuleEntity validationFunction = ModuleEntity.wrap(bytes24(userOp.signature[:24]));
+        uint8 validationFlags = uint8(userOp.signature[24]);
+        bool isGlobalValidation = validationFlags & 1 == 1;
+        bool isDeferredInstallValidation = validationFlags & 2 == 2;
```
adamegyed: nit: could we define the `1` and `2` here as `uint8` constants, and use `!= 0` as the checking condition instead of the `== <constant>` form?

src/account/ModularAccount.sol
```diff
@@ -352,32 +366,139 @@ contract ModularAccount is
...
+            // deferred validation.
+            bytes calldata deferredInstallSig =
+                userOp.signature[33 + encodedDataLength:33 + encodedDataLength + deferredInstallSigLength];
+
+            //Validate the signature.
+            _validateDeferredInstallDataAndSetNonce(
+                validationFunction, deferredValidationInstallData, deferredInstallSig
+            );
+
+            // Use a self-call to install the deferred validation.
```
adamegyed, line 437: question: can we use the internal `_installValidation`, or do we need to convert calldata types to memory types?

src/account/ModularAccount.sol
```diff
@@ -352,32 +366,139 @@ contract ModularAccount is
...
+
+            // Signatures need to stay in calldata, so we decode them as if they were two concatenated byte arrays.
+            // Get the deferred installation signature length.
+            uint256 deferredInstallSigLength =
+                uint32(bytes4(userOp.signature[29 + encodedDataLength:33 + encodedDataLength]));
+
+            // Get the deferred installation signature, which is passed to the outer validation to install the
+            // deferred validation.
+            bytes calldata deferredInstallSig =
+                userOp.signature[33 + encodedDataLength:33 + encodedDataLength + deferredInstallSigLength];
```
adamegyed, line 430: I think all of the offsets here are correct. For readability, could you add a comment describing the components of `userOp.signature` and how long each is? Like the comment in the test, or in another format if it works better.

adamegyed (commented):
Awesome feature 🔥 

Had a few small comments

src/account/ModularAccount.sol
Zer0dot: There's no actual benefit-- I just do this by convention, but making it public does save a little bit of bytecode! Happy to do that.

src/account/ModularAccount.sol
Zer0dot: Sure, makes sense!

src/account/ModularAccount.sol
Zer0dot: so `_installValidation()` does do some calldata bytes slicing. It's possible to modify it to work internally in memory, but we might have efficiency loss in the standard `installValidation()` case. My thoughts were to just keep it simple and run a self-call. Wdyt?

src/account/ModularAccount.sol
Zer0dot: Yes absolutely, this is the part of this PR that's the most confusing. Will write a comment to clarify this.

src/account/ModularAccount.sol
Zer0dot: Done in [a3a4467](https://github.com/alchemyplatform/modular-account/pull/177/commits/a3a44679f4a3f2e0b9db6c482d9ebd11d23bf46b)!

src/account/ModularAccount.sol
Zer0dot: Done in [a3a4467](https://github.com/alchemyplatform/modular-account/pull/177/commits/a3a44679f4a3f2e0b9db6c482d9ebd11d23bf46b)!

src/account/ModularAccount.sol
Zer0dot: Added in [a3a4467](https://github.com/alchemyplatform/modular-account/pull/177/commits/a3a44679f4a3f2e0b9db6c482d9ebd11d23bf46b)!

src/account/ModularAccount.sol
adamegyed: Ah yep, that's what I thought. The current approach makes sense, no need to do big refactors at the moment imo.

src/account/ModularAccount.sol
```diff
@@ -352,32 +369,151 @@ contract ModularAccount is
         override
         returns (uint256 validationData)
     {
+        // Use outer validation as 1271 validation, then installed validation to validate the rest
```
adamegyed: tiny tiny nit: I think this comment only applies in the case where we're using the deferred validation install path, so could you move this comment to within the `if (isDeferredInstallValidation) {` block?

adamegyed (approved):
Looks good, 1. small nit

src/account/ModularAccount.sol
Zer0dot: Done!

PR #178, opened 2024-09-16
feat: add allowlist module

adamegyed:
## Motivation

We want the ability to apply an allowlist to specific validations on an account.

## Solution

Port over the Allowlist module from the RI. This checks addresses and optionally also checks selectors. It does not check any additional portions of calldata.

### DIFF
--- a/src/modules/permissions/AllowlistModule.sol
+++ b/src/modules/permissions/AllowlistModule.sol
@@ -1,0 +1,194 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+import {IERC165} from "@openzeppelin/contracts/interfaces/IERC165.sol";
+
+import {Call, IModularAccount} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
+import {IModule} from "@erc6900/reference-implementation/interfaces/IModule.sol";
+import {IValidationHookModule} from "@erc6900/reference-implementation/interfaces/IValidationHookModule.sol";
+
+import {BaseModule} from "../../modules/BaseModule.sol";
+
+/// @title Allowlist Module
+/// @author Alchemy
+/// @notice This module allows for the setting and enforcement of allowlists for validation functions. These
+/// allowlists can specify which addresses can be called by the validation function, and optionally which selectors
+/// can be called on those addresses. These restrictions only apply to the `IModularAccount.execute` and
+/// `IModularAccount.executeBatch` functions.
+contract AllowlistModule is IValidationHookModule, BaseModule {
+    struct AllowlistInit {
+        address target;
+        bool hasSelectorAllowlist;
+        bytes4[] selectors;
+    }
+
+    struct AllowlistEntry {
+        bool allowed;
+        bool hasSelectorAllowlist;
+    }
+
+    mapping(uint32 entityId => mapping(address target => mapping(address account => AllowlistEntry))) public
+        targetAllowlist;
+    mapping(
+        uint32 entityId => mapping(address target => mapping(bytes4 selector => mapping(address account => bool)))
+    ) public selectorAllowlist;
+
+    event AllowlistTargetUpdated(
+        uint32 indexed entityId, address indexed account, address indexed target, AllowlistEntry entry
+    );
+    event AllowlistSelectorUpdated(
+        uint32 indexed entityId, address indexed account, bytes24 indexed targetAndSelector, bool allowed
+    );
+
+    error TargetNotAllowed();
+    error SelectorNotAllowed();
+    error NoSelectorSpecified();
+
+    /// @inheritdoc IModule
+    /// @dev The `data` parameter is expected to be encoded as `(uint32 entityId, AllowlistInit[] init)`.
+    function onInstall(bytes calldata data) external override {
+        (uint32 entityId, AllowlistInit[] memory init) = abi.decode(data, (uint32, AllowlistInit[]));
+
+        for (uint256 i = 0; i < init.length; i++) {
+            setAllowlistTarget(entityId, init[i].target, true, init[i].hasSelectorAllowlist);
+
+            if (init[i].hasSelectorAllowlist) {
+                for (uint256 j = 0; j < init[i].selectors.length; j++) {
+                    setAllowlistSelector(entityId, init[i].target, init[i].selectors[j], true);
+                }
+            }
+        }
+    }
+
+    /// @inheritdoc IModule
+    /// @dev The `data` parameter is expected to be encoded as `(uint32 entityId, AllowlistInit[] init)`.
+    function onUninstall(bytes calldata data) external override {
+        (uint32 entityId, AllowlistInit[] memory init) = abi.decode(data, (uint32, AllowlistInit[]));
+
+        batchInitAllowlist(entityId, init);
+    }
+
+    /// @inheritdoc IValidationHookModule
+    function preUserOpValidationHook(uint32 entityId, PackedUserOperation calldata userOp, bytes32)
+        external
+        view
+        override
+        returns (uint256)
+    {
+        checkAllowlistCalldata(entityId, userOp.callData);
+        return 0;
+    }
+
+    /// @inheritdoc IValidationHookModule
+    function preRuntimeValidationHook(uint32 entityId, address, uint256, bytes calldata data, bytes calldata)
+        external
+        view
+        override
+    {
+        checkAllowlistCalldata(entityId, data);
+        return;
+    }
+
+    // solhint-disable-next-line no-empty-blocks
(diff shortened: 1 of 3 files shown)


src/modules/permissions/AllowlistModule.sol
```diff
@@ -0,0 +1,155 @@
...
+    struct AllowlistInit {
+        address target;
+        bool hasSelectorAllowlist;
+        bytes4[] selectors;
+    }
+
+    struct AllowlistEntry {
+        bool allowed;
+        bool hasSelectorAllowlist;
+    }
```
fangting-alchemy, line 29: I see what you are doing. Address is used as the entry point for both addresses and selectors. It wont support cases where user wants to keep addresses and selectors limits separate.

How about something like this?
```
    mapping(bytes4 selector => mapping(address target => mapping(address account => bool))))
    ) public selectorAllowList;
     mapping(address target => mapping(bytes4 selector => mapping(address account => bool))))
    ) public contractAllowList;
```
For wildcard support, we set the wildcard to `address(0)` or `bytes4('')`.

When checking, we check the following:
```
selectorAllowList[selector][target] ||
selectorAllowList[selector][address(0)] ||
contractAllowList[target][bytes4('')]
```
We can avoid `contractAllowList[target][selector]` by use `selectorAllowList` as priority.

adamegyed: That could work, but I think it would increase the checking costs in most cases because it needs to cover 2 cases sequentially.

adamegyed: I'll implement it for now, and optimize later if/when we combine the checking hooks.

adamegyed: oh wait, this doesn't work - `bytes4(0)` is a real value, we can't treat it as a wildcard.

Let's keep it as-is for now, and move the address-wildcard restriction to a different PR.

fangting-alchemy: It would require a major refactor almost like a rewrite. Probably slightly easier to review as a new file.

Why cannot we treat `bytes4(uint32(0))` as a wildcard?

adamegyed: > Why cannot we treat bytes4(uint32(0)) as a wildcard?

That is a valid selector - some contracts mine an all-zero selector, see https://www.4byte.directory/signatures/?bytes4_signature=0x00000000

If someone adds that selector to the allowlist, they would accidentally be adding a wildcard, and it may not be obvious if they're just retrieving things from an ABI.

fangting-alchemy: TIL.

We can update the second mapping actually. Better anyways.

```
mapping(bytes4 selector => mapping(address target => mapping(address account => bool))))
    ) public selectorAllowList;
mapping(address target => mapping(address account => bool))) public contractAllowList;
```
The second mapping is now basically the wildcard for selector for the contract, since all non-wildcard cases can be handled by the first mapping.

src/modules/permissions/AllowlistModule.sol
```diff
@@ -0,0 +1,155 @@
...
+
+    // solhint-disable-next-line no-empty-blocks
+    function preSignatureValidationHook(uint32, address, bytes32, bytes calldata) external pure override {}
+
+    /// @inheritdoc IModule
+    function moduleId() external pure returns (string memory) {
+        return "alchemy.allowlist-module.0.0.1";
+    }
+
+    function setAllowlistTarget(uint32 entityId, address target, bool allowed, bool hasSelectorAllowlist) public {
```
fangting-alchemy, line 123: For a target with selectors, would two calls required to set it up? Should we add one more input here to accept optional selectors?

fangting-alchemy: What we have in `oninstall` is does take that into consideration. We probably can reuse the logic there.

src/modules/permissions/AllowlistModule.sol
```diff
@@ -0,0 +1,155 @@
...
+        AllowlistEntry memory entry = AllowlistEntry(allowed, hasSelectorAllowlist);
+        targetAllowlist[entityId][target][msg.sender] = entry;
+        emit AllowlistTargetUpdated(entityId, msg.sender, target, entry);
+    }
+
+    function setAllowlistSelector(uint32 entityId, address target, bytes4 selector, bool allowed) public {
+        selectorAllowlist[entityId][target][selector][msg.sender] = allowed;
```
fangting-alchemy, line 137: When this is set for the first time for a target, should we update the `hasSelectorAllowlist` value in`targetAllowlist`?

src/modules/permissions/AllowlistModule.sol
adamegyed: Yeah, I think right now you would need to batch multiple steps together. _Technically_, the accounts can now directly call `onInstall` even after the module is already initialized, so they could use it to batch the call, but that's a bit of a clumsy workflow. Do you think we should add a new external function to do the same workflow as `onInstall`, just named something else? It'd be more intuitive for devs.

src/modules/permissions/AllowlistModule.sol
adamegyed: I think we can't tell whether or not it's the first target, because these mappings are non-enumerable.

src/modules/permissions/AllowlistModule.sol
fangting-alchemy: > Do you think we should add a new external function to do the same workflow as onInstall, just named something else? It'd be more intuitive for devs.

Yeah, that'd be good. They can call the same internal func under the hood.

src/modules/permissions/AllowlistModule.sol
fangting-alchemy: In that case, should we do sth like the following?
```
If (allowed) {
   intargetAllowlist[entityId][target].hasSelectorAllowlist = true;
}
```

Otherwise, we will have to revert selector direct set for target where the target has `hasSelectorAllowlist` set to true. Because in this case, this selector wont be checked. So users will have to add selectors through the other set method.

I like the first option better.

src/modules/permissions/AllowlistModule.sol
adamegyed: Added!

src/modules/permissions/AllowlistModule.sol
adamegyed: I think we should avoid those extra storage writes - the user can already specify these two things independently. To address this, I kept the individual edit functions (`setAllowlistTarget`, `setAllowlistSelector`) in, as well as the new batch init function `batchInitAllowlist`.

fangting-alchemy: The case I am concerned is.

- User installs hook to limit a SK to call a contract (`addressA`).
- User realizes that they can only allow one selector on `addressA` for the SK to achieve whatever they want. So they come to call `setAllowlistSelector`, and think that it would work.
- It wont work because `addressA` has `hasSelectorAllowlist` to be true.

So I think this storage writes is actually worth it in this case.

The only alternative is to educate users/customers to be aware of this, which seems it will be more work.

adamegyed: I think if they're on the level of calling individual functions on this contract, rather than using some higher-level abstractions from the SDK, then they need to be comfortable with understanding how the contract works. We shouldn't hamper performance to address a hypothetical user error.

fangting-alchemy: >We shouldn't hamper performance to address a hypothetical user error.

The only performance weakness is `If (allowed) {` this portion. Which is almost nothing.
- if `intargetAllowlist[entityId][target].hasSelectorAllowlist` was true, the set costs nothing.
- if it was originally false, (without setting it to true) the whole method call is wasted as it won't do anything.

adamegyed: > the set costs nothing.

This is not true, solidity loads every storage slot before setting to keep other bits the same, and even a warm load or warm set costs something. ([ref](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html#:~:text=When%20one%20value%20is%20written%20to%20a%20multi%2Dvalue%20storage%20slot%2C%20the%20storage%20slot%20has%20to%20be%20read%20first%20and%20then%20combined%20with%20the%20new%20value%20such%20that%20other%20data%20in%20the%20same%20slot%20is%20not%20destroyed.))

The users sets the config as they wish, including the ability to set it to something that isn't valid or useful. It's not the contract's job to enforce that, that happens on a different level.

fangting-alchemy: Oh yeah, missed the loading part. Hmm. Need to weigh more.

Anyways, to no block this PR, can you add this behavior to docs as dev notice to warn this case?

adamegyed: Added a comment to call this out in the natspec, to warn users that this setting will only matter if the target address has the selector allowlist enabled.

fangting-alchemy (approved):

PR #179, opened 2024-09-16
feat: add MA runtime gas benchmarks

adamegyed:
## Motivation

Continue work on building out benchmarking for MAv2, this time covering two new runtime cases: native token transfer, and ERC-20 transfer.

## Solution

Add a runtime gas benchmark for native token transfer and ERC-20 transfer for both ModularAccount and LightAccount.

The LightAccount test required importing the LightAccount ABI as an interface.

Verified that the benchmark is accurate by comparing the LightAccount v2 benchmark to its corresponding result in [aa-benchmarks](https://github.com/alchemyplatform/aa-benchmarks).

Unify the remappings setup for the gas tests, so now the entire repo uses the convention from `forge-gas-snapshot`. Without this, it wouldn't be possible to import the test utilities from `test/` into `gas/`.

Move the 6900 signature encoding test utility functions out to a separate file `ModuleSignatureUtils.sol`, so it can be imported in the gas tests without needing to extend the test setup `AccountTestBase`.

Add common workflows for testing, generating gas reports, and all pre-submission prep to `package.json`.


fangting-alchemy (approved):

remappings.txt
```diff
@@ -6,4 +6,5 @@ account-abstraction/=node_modules/account-abstraction/contracts/
 @alchemy/light-account/=lib/light-account/
 solady=node_modules/solady/src/
 @erc-6900/reference-implementation/=node_modules/reference-implementation/src/
-forge-gas-snapshot/=lib/forge-gas-snapshot/src/
\ No newline at end of file
+forge-gas-snapshot/=lib/forge-gas-snapshot/src/
+forge-std/=lib/forge-std/
```
jaypaik, line 10: @adamegyed was there a reason `src` was omitted here?

adamegyed: This was the way that [forge-gas-snapshot](https://github.com/marktoda/forge-gas-snapshot) did its import, so if we kept it as before, we couldn't compile the test files in `gas/`. In a previous PR [here](https://github.com/alchemyplatform/modular-account/pull/176/files#diff-bac743fe3d899998e32393af53af8cc7d28c89c3d7fabfa48b01361cf8d42d9b), I tried to have two different remappings setups, but it became too unwieldy when trying to import files from `test/utils/` into `gas/`, so in this PR I changed the entire repo to use the import scheme from forge-gas-snapshot.

adamegyed: That said, since we integrated forge-gas-snapshot, its features have been upstreamed into foundry itself, so we could migrate things over to remove the external dependency and change back to the old remappings layout. It'll just take a bit of porting work.

PR #180, opened 2024-09-18
feat: add time range module

adamegyed:
## Motivation

We want the ability to limit validations by time range, for use in session keys and other modules.

## Solution

Add a time range enforcing module. When used as a user op validation, this returns the time range in the 4337 validationData value. When used as a runtime validation, this checks `block.timestamp` and enforces that the validation is only used within this context.

Also adds additional commands to `package.json` for viewing coverage and generating an `lcov` report.

Note: this currently doesn't enforce anything for signature validation. It is not clear to me whether or not it is safe to use `block.timestamp` in the signature checking logic, because if may be used within another account's 4337 validation phase. However, that entire workflow is not common due to the likely violation for associated storage.

To discuss: should this hook check `block.timestamp` in the pre signature validation hook? Or should we just recommend users to disable signature validation from validation functions with this hook using the flag?

### DIFF
--- a/src/modules/permissions/TimeRangeModule.sol
+++ b/src/modules/permissions/TimeRangeModule.sol
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {_packValidationData} from "@eth-infinitism/account-abstraction/core/Helpers.sol";
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+import {IERC165} from "@openzeppelin/contracts/interfaces/IERC165.sol";
+
+import {IModule} from "@erc6900/reference-implementation/interfaces/IModule.sol";
+import {IValidationHookModule} from "@erc6900/reference-implementation/interfaces/IValidationHookModule.sol";
+
+import {BaseModule} from "../../modules/BaseModule.sol";
+
+/// @title Time Range Module
+/// @author Alchemy
+/// @notice This module allows for the setting and enforcement of time ranges for a validation function. Enforcement
+/// relies on `block.timestamp`, either within this module for runtime validation, or by the EntryPoint for user op
+/// validation. Time ranges are inclusive of both the beginning and ending timestamps.
+contract TimeRangeModule is IValidationHookModule, BaseModule {
+    struct TimeRange {
+        uint48 validUntil;
+        uint48 validAfter;
+    }
+
+    mapping(uint32 entityId => mapping(address account => TimeRange)) public timeRanges;
+
+    error TimeRangeNotValid();
+
+    /// @inheritdoc IModule
+    /// @notice Initializes the module with the given time range for `msg.sender` with a given entity id.
+    /// @dev data is abi-encoded as (uint32 entityId, uint48 validUntil, uint48 validAfter)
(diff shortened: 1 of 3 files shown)


test/modules/TimeRangeModule.t.sol
```diff
@@ -0,0 +1,204 @@
...
+
+        // Verify that the time range data is unset
+        (uint48 retrievedValidUntil, uint48 retrievedValidAfter) =
+            timeRangeModule.timeRanges(HOOK_ENTITY_ID, address(account1));
+
+        assertEq(retrievedValidUntil, 0);
+        assertEq(retrievedValidAfter, 0);
+    }
+
+    function testFuzz_timeRangeModule_userOp(uint48 _validUntil, uint48 _validAfter) public {
```
fangting-alchemy, line 89: Should we add a failure case for this path?

adamegyed: What does a failure case test look like? Once it is uninstalled, the account won't make any calls towards the module anymore.

fangting-alchemy: I meant a case fore sigFailed to be true.

adamegyed: Ah I see, adding...

fangting-alchemy (approved):

src/modules/permissions/TimeRangeModule.sol
```diff
@@ -0,0 +1,82 @@
...
+import {_packValidationData} from "@eth-infinitism/account-abstraction/core/Helpers.sol";
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+import {IERC165} from "@openzeppelin/contracts/interfaces/IERC165.sol";
+
+import {IModule} from "@erc6900/reference-implementation/interfaces/IModule.sol";
+import {IValidationHookModule} from "@erc6900/reference-implementation/interfaces/IValidationHookModule.sol";
+
+import {BaseModule} from "../../modules/BaseModule.sol";
+
+contract TimeRangeModule is IValidationHookModule, BaseModule {
```
jaypaik, line 18: Mind adding NatSpec documentation to our contracts?

src/modules/permissions/TimeRangeModule.sol
```diff
@@ -0,0 +1,82 @@
...
+    struct TimeRange {
+        uint48 validUntil;
+        uint48 validAfter;
+    }
+
+    mapping(uint32 entityId => mapping(address account => TimeRange)) public timeRanges;
+
+    error TimeRangeNotValid();
+
+    function onInstall(bytes calldata data) external override {
```
jaypaik, line 31: Also `@dev` comments in functions like these to describe how data should be encoded.

src/modules/permissions/TimeRangeModule.sol
adamegyed: Added here: https://github.com/alchemyplatform/modular-account/pull/180/commits/5aeff4d4057d613831e4577f70a0a3c58b68c076

Lmk if anything's missing!

src/modules/permissions/TimeRangeModule.sol
adamegyed: Added!

Though I wonder if it would be preferable to put this info in an `@param` instead?

PR #181, opened 2024-09-19
feat: user op gas benchmarks for MA and LA

adamegyed:
## Motivation

We want benchmarks for using accounts in the 4337 usage path, via a user op.

## Solution

Add user op versions of the "native token transfer" and "erc-20 transfer" gas benchmark cases for LightAccount v2 and for ModularAccount. There is not yet a test case for the SemiModularAccount.

There is also no test case for deploying an account in a user op, because it is hard to separate the gas costs of other steps here. For now, I think the runtime account creation case is sufficient.

Also refactors some parts of the benchmark base out from the ModularAccount-specific components, to be shared by the benchmarks for different account types, and renames files accordingly.

### Change from original gas benchmarking plan

in the gas benchmarking plan doc, we wanted to use the `UserOperationEvent`s emitted by the EntryPoint to get more accurate gas measurements, rather than just measuring the gas cost of a 1 user op call to `handleOps`. However, this poses some additional challenges:
- We need to estimate the gas limits used by user ops, because if we just apply a very large limit, the the EntryPoint will assess a [penalty](https://github.com/eth-infinitism/account-abstraction/blob/v0.7.0/contracts/core/EntryPoint.sol#L718). While possible, it is not straightforward to recreate `rundler`'s gas estimation purely in foundry.
- We also need an estimate on `preVerificationGas`. If we just set this to zero, the gas used by `handleOps` will exceed the gas that the EntryPoint attributes to the user op. Additionally, it is not straightforward to determine the correct PVG - it changes by L1/L2 network, and thus isn't as simple as the intrinsic txn cost + calldata costs.

To keep things simple and matching what we're currently doing in aa-benchmarks, these tests just measure the execution gas usage of the call to `handleOps` with a single user op, performing the action in the test case. We can revisit this in the future, if we're comfortable handling the issues mentioned above.


gas/modular-account/ModularAccount.gas.t.sol
```diff
@@ -0,0 +1,152 @@
...
+
+        VmSafe.Gas memory gas = vm.lastCallGas();
+
+        console.log("Runtime: account creation: ");
+        console.log("gasTotalUsed: %d", gas.gasTotalUsed);
+
+        snap("ModularAccount_Runtime_AccountCreation", gas.gasTotalUsed);
+    }
+
+    function test_modularAccountGas_runtime_nativeTransfer() public {
```
Zer0dot, line 29: This is a tangent, but do you think we should also benchmark the direct call path?

adamegyed: Ah yeah, probably worth benchmarking eventually, too. It's not super high-priority because it'll mostly be used by modules, though.

adamegyed: Additional side-note: it would be really nice if SMA interprets direct calls from its owner as if it passed runtime validation, similar to the direct-call validation path.

Zer0dot: So a way to bypass `executeWithAuthorization()` and just directly call whatever function? Sounds reasonable!

gas/modular-account/ModularAccountBenchmarkBase.sol
```diff
@@ -12,40 +10,23 @@ import {SemiModularAccount} from "../../src/account/SemiModularAccount.sol";
...
-import {MockERC20} from "../../test/mocks/MockERC20.sol";
-
 import {ModuleSignatureUtils} from "../../test/utils/ModuleSignatureUtils.sol";
-import {OptimizedTest} from "../../test/utils/OptimizedTest.sol";
-
-abstract contract ModularAccountBenchmarkBase is GasSnapshot, OptimizedTest, ModuleSignatureUtils {
-    EntryPoint public entryPoint;
+import {BenchmarkBase} from "..//BenchmarkBase.sol";
 
+abstract contract ModularAccountBenchmarkBase is BenchmarkBase, ModuleSignatureUtils {
```
Zer0dot, line 16: This looks like it's only inherited once-- if I understand right, the expectation is that we'll use this for other tests too, right?

Zer0dot (approved):
No comments besides some small informational stuff, lgtm!

gas/modular-account/ModularAccountBenchmarkBase.sol
adamegyed: Yep, see https://github.com/alchemyplatform/modular-account/pull/183!

PR #182, opened 2024-09-20
fix: fix how 6900 dependencies are mapped and imported

fangting-alchemy:
## Motivation

The `modular-account-libs` were linked unconventionally.

## Solution
Add `@` to remapping for `modular-account-libs`.

### DIFF
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,5 +1,5 @@
 ds-test/=lib/forge-std/lib/ds-test/src/
-@modular-account-libs/=node_modules/modular-account-libs/src/
+@erc6900/modular-account-libs/=node_modules/modular-account-libs/src/
 @eth-infinitism/account-abstraction/=node_modules/account-abstraction/contracts/
 account-abstraction/=node_modules/account-abstraction/contracts/
 @openzeppelin/=lib/openzeppelin-contracts/
--- a/remappings.txt
+++ b/remappings.txt
@@ -1,5 +1,5 @@
 ds-test/=lib/forge-std/lib/ds-test/src/
-modular-account-libs/=node_modules/modular-account-libs/src/
+@modular-account-libs/=node_modules/modular-account-libs/src/
 @eth-infinitism/account-abstraction/=node_modules/account-abstraction/contracts/
 account-abstraction/=node_modules/account-abstraction/contracts/
 @openzeppelin/=lib/openzeppelin-contracts/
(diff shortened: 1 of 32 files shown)


remappings.txt
```diff
@@ -1,5 +1,5 @@
 ds-test/=lib/forge-std/lib/ds-test/src/
-modular-account-libs/=node_modules/modular-account-libs/src/
+@modular-account-libs/=node_modules/modular-account-libs/src/
```
jaypaik: Can we do:
```suggestion
@erc6900/modular-account-libs/=node_modules/@erc6900/modular-account-libs/src/
```

And update package.json to follow the convention used here: https://github.com/erc6900/modular-account-libs/blob/d64adb5bd100d6ad64d0a83b5dc13b356423d852/package.json#L5

adamegyed: nit: If we use `@erc6900/` here, could we update the reference impl import, too? Currently that one uses `@erc-6900`, which is not the GitHub org name.

fangting-alchemy: Sounds good on adding `@erc6900`. Will adda backlog for RI.

@jaypaik Why `node_modules/@erc6900/modular-account-libs/src/`? we will need to update the published NPM module.

adamegyed: @fangting-alchemy it is not actually published as an NPM module, we just use the `github:` syntax to import it in `package.json`: https://github.com/alchemyplatform/modular-account/blob/develop/package.json#L6

To use this workflow, the repo just needs a `package.json` in the repo root with a `name` and `version` field.

remappings.txt
```diff
@@ -1,10 +1,10 @@
 ds-test/=lib/forge-std/lib/ds-test/src/
-modular-account-libs/=node_modules/modular-account-libs/src/
+@erc6900/modular-account-libs/=node_modules/@erc6900/modular-account-libs/src/
 @eth-infinitism/account-abstraction/=node_modules/account-abstraction/contracts/
 account-abstraction/=node_modules/account-abstraction/contracts/
 @openzeppelin/=lib/openzeppelin-contracts/
 @alchemy/light-account/=lib/light-account/
 solady=node_modules/solady/src/
-@erc-6900/reference-implementation/=node_modules/reference-implementation/src/
+@erc-6900/reference-implementation/=node_modules/@erc6900/reference-implementation/src/
```
jaypaik:
```suggestion
@erc6900/reference-implementation/=node_modules/@erc6900/reference-implementation/src/
```

jaypaik (approved):
1 nit! Thanks for doing this!

PR #183, opened 2024-09-20
feat: benchmark SemiModularAcocunt

adamegyed:
## Motivation

We also want accurate gas benchmarks for the SemiModularAccount.

## Solution

Refactor the BenchmarkBase for more code reuse and less duplication, and update ModularAccount and LightAccount gas benchmarks accordingly.

Add a gas benchmark test for SemiModularAccount. All of these cases currently use the original bytecode-appended owner, will need additional benchmarks for testing cases with a rotated owner in storage as a follow-up.

Refactors SMA to have the constant for the fallback signer pulled in from `Constants.sol`, rather than an internal constant, for easier use in tests.


gas/BenchmarkBase.sol
```diff
@@ -32,6 +45,44 @@ abstract contract BenchmarkBase is OptimizedTest, GasSnapshot {
...
+    }
+
+    function _snap(BenchmarkType bType, string memory testCase, uint256 gasValue) internal {
+        string memory consoleLine = string.concat(_benchmarkTypeToString(bType), ": ", testCase, ": ");
+
+        console.log(consoleLine);
+        console.log("gasTotalUsed: %d", gasValue);
+
+        string memory snapName =
+            string.concat(string(_accountImplName), "_", _benchmarkTypeToString(bType), "_", testCase);
```
Zer0dot: is the `string()` cast necessary here? `_accountImplName` is already a string-- but this could also be a Solidity quirk.

gas/modular-account/SemiModularAccount.gas.t.sol
```diff
@@ -0,0 +1,133 @@
...
+
+contract ModularAccountGasTest is ModularAccountBenchmarkBase("SemiModularAccount") {
+    function test_semiModularAccountGas_runtime_accountCreation() public {
+        uint256 salt = 0;
+
+        uint256 gasUsed = _runtimeBenchmark(
+            owner1, address(factory), abi.encodeCall(factory.createSemiModularAccount, (owner1, salt))
+        );
+
+        assertTrue(factory.getAddressSemiModular(owner1, salt).code.length > 0);
```
Zer0dot: Mega small nit but depending on the context this check might not guarantee that the contract was deployed-- maybe we could check the factory event to ensure it was deployed?

Zer0dot: (Same for the standard ModularAccount gas test)

Zer0dot (approved):
Great refactor here too, this makes it a lot easier to reuse snapshotting logic. Thanks for all this, no other comments!

gas/BenchmarkBase.sol
adamegyed: Nope it's not necessary, I left it here accidentally. Fixed!

gas/modular-account/SemiModularAccount.gas.t.sol
adamegyed: Is the issue you're thinking about that the address returned by `getAddress(SemiModular)?` might be a different address than what the factory actually deploys to? Or something else?

I added the event check, and also kept this code size check.

PR #184, opened 2024-09-20
feat: add erc20 token limit module

fangting-alchemy:
## Motivation

Account holders may want to set ERC20 token spend limit to certain validations. 

## Solution
Create an ERC20 token limit module to offer hooks associated with validations that works with both UO and runtime call paths. See code doc for features and restrictions.

### DIFF
--- a/src/modules/ERC20TokenLimitModule.sol
+++ b/src/modules/ERC20TokenLimitModule.sol
@@ -2,7 +2,6 @@
 pragma solidity ^0.8.26;
 
 import {UserOperationLib} from "@eth-infinitism/account-abstraction/core/UserOperationLib.sol";
-import {IAccountExecute} from "@eth-infinitism/account-abstraction/interfaces/IAccountExecute.sol";
 import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
 
 import {
@@ -30,8 +29,7 @@ import {BaseModule, IERC165} from "./BaseModule.sol";
 /// IModularAccount.executeBatch. All other spending request will be reverted.
 ///     - this module is opinionated on what selectors (transfer and approve only) can be called for token
 /// contracts to guard against weird edge cases like DAI. You wouldn't be able to use uni v2 pairs directly as the
-/// pair contract is also the LP token contract
-
+/// pair contract is also the LP token contract.
 contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
     using UserOperationLib for PackedUserOperation;
     using EnumerableSet for EnumerableSet.AddressSet;
@@ -42,21 +40,20 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
         uint256 limit;
     }
 
-    string internal constant _NAME = "ERC20 Token Limit Module";
-    string internal constant _VERSION = "1.0.0";
-    string internal constant _AUTHOR = "Alchemy & ERC-6900 Authors";
-
     mapping(uint32 entityId => mapping(address token => mapping(address account => uint256 limit))) public limits;
     AssociatedLinkedListSet internal _tokenList;
 
     error ExceededTokenLimit();
-    error ExceededNumberOfEntities();
     error SelectorNotAllowed();
     error SpendingRequestNotAllowed(bytes4);
     error ERC20NotAllowed(address);
+    error InvalidCalldataLength();
 
     function updateLimits(uint32 entityId, address token, uint256 newLimit) external {
-        _tokenList.tryAdd(msg.sender, SetValue.wrap(bytes30(bytes20(token))));
+        if (token == address(0)) {
+            revert ERC20NotAllowed(address(0));
+        }
+        _tokenList.tryAdd(msg.sender, _toSetValue(token));
         limits[entityId][token][msg.sender] = newLimit;
     }
 
@@ -68,14 +65,10 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
     {
         (bytes4 selector, bytes memory callData) = _getSelectorAndCalldata(data);
 
-        if (
-            selector == IModularAccount.execute.selector
-                || selector == IModularAccount.executeWithAuthorization.selector
-                || selector == IAccountExecute.executeUserOp.selector
-        ) {
+        if (selector == IModularAccount.execute.selector) {
             // when calling execute or ERC20 functions directly
             (address token,, bytes memory innerCalldata) = abi.decode(callData, (address, uint256, bytes));
-            if (_tokenList.contains(msg.sender, SetValue.wrap(bytes30(bytes20(token))))) {
+            if (_tokenList.contains(msg.sender, _toSetValue(token))) {
                 _decrementLimit(entityId, token, innerCalldata);
             } else {
                 revert ERC20NotAllowed(token);
@@ -83,7 +76,7 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
         } else if (selector == IModularAccount.executeBatch.selector) {
             Call[] memory calls = abi.decode(callData, (Call[]));
             for (uint256 i = 0; i < calls.length; i++) {
-                if (_tokenList.contains(msg.sender, SetValue.wrap(bytes30(bytes20(calls[i].target))))) {
+                if (_tokenList.contains(msg.sender, _toSetValue(calls[i].target))) {
                     _decrementLimit(entityId, calls[i].target, calls[i].data);
                 } else {
                     revert ERC20NotAllowed(calls[i].target);
@@ -100,12 +93,18 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
         (uint32 entityId, ERC20SpendLimit[] memory spendLimits) = abi.decode(data, (uint32, ERC20SpendLimit[]));
 
         for (uint8 i = 0; i < spendLimits.length; i++) {
-            _tokenList.tryAdd(msg.sender, SetValue.wrap(bytes30(bytes20(spendLimits[i].token))));
-            limits[entityId][spendLimits[i].token][msg.sender] = spendLimits[i].limit;
+            address token = spendLimits[i].token;
+            if (token == address(0)) {
+                revert ERC20NotAllowed(address(0));
+            }
+            _tokenList.tryAdd(msg.sender, _toSetValue(token));
+            limits[entityId][token][msg.sender] = spendLimits[i].limit;
         }
     }
 
     /// @inheritdoc IModule
+    /// @notice uninstall this module can only clear limit for one token of one entity. To clear all limits, users
+    /// are recommended to use updateLimit for each token and entityId.
     function onUninstall(bytes calldata data) external override {
         (address token, uint32 entityId) = abi.decode(data, (address, uint32));
         delete limits[entityId][token][msg.sender];
@@ -127,7 +126,7 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
 
     /// @inheritdoc IModule
     function moduleId() external pure returns (string memory) {
-        return "erc6900.erc20-token-limit-module.1.0.0";
+        return "alchemy.erc20-token-limit-module.0.0.1";
     }
 
     /// @inheritdoc BaseModule
@@ -136,18 +135,21 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
     }
 
     function _decrementLimit(uint32 entityId, address token, bytes memory innerCalldata) internal {
+        if (innerCalldata.length < 68) {
+            revert InvalidCalldataLength();
+        }
+
         bytes4 selector;
         uint256 spend;
         assembly {
             selector := mload(add(innerCalldata, 32)) // 0:32 is arr len, 32:36 is selector
             spend := mload(add(innerCalldata, 68)) // 36:68 is recipient, 68:100 is spend
         }
-        if (selector == IERC20.transfer.selector || selector == IERC20.approve.selector) {
+        if (_isAllowedERC20Function(selector)) {
             uint256 limit = limits[entityId][token][msg.sender];
             if (spend > limit) {
                 revert ExceededTokenLimit();
             }
-            // solhint-disable-next-line reentrancy
             limits[entityId][token][msg.sender] = limit - spend;
         } else {
             revert SelectorNotAllowed();
@@ -156,5 +158,9 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
 
     function _isAllowedERC20Function(bytes4 selector) internal pure returns (bool) {
         return selector == IERC20.transfer.selector || selector == IERC20.approve.selector;
+    }
+
+    function _toSetValue(address token) internal pure returns (SetValue) {
+        return SetValue.wrap(bytes30(bytes20(token)));
     }
 }
--- a/src/modules/ERC20TokenLimitModule.sol
+++ b/src/modules/ERC20TokenLimitModule.sol
@@ -1,0 +1,159 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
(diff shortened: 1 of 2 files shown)


src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+        external
+        override
+        returns (bytes memory)
+    {
+        (bytes4 selector, bytes memory callData) = _getSelectorAndCalldata(data);
+
+        if (
+            selector == IModularAccount.execute.selector
+                || selector == IModularAccount.executeWithAuthorization.selector
+                || selector == IAccountExecute.executeUserOp.selector
```
adamegyed: I think the split-path case between the regular `execute` and `executeUserOp` is handled by is handled by `_getSelectorAndCalldata` already, so adding `executeUserOp` isn't necessary here.

Additionally, I think the account would never fire out a hook with `executeWIthAuthorization` as the selector, since that function doesn't have the `wrapNativeFunction` modifier. We should double check that call path, but I think it is ok to exclude here (even if it were sent here, the abi decoding on the next line would fail because it is encoded differently).

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /// @inheritdoc IExecutionHookModule
+    function postExecutionHook(uint32, bytes calldata) external pure override {
+        revert NotImplemented();
+    }
+
+    /// @inheritdoc IModule
+    function moduleId() external pure returns (string memory) {
+        return "erc6900.erc20-token-limit-module.1.0.0";
```
adamegyed:
```suggestion
        return "alchemy.erc20-token-limit-module.0.0.1";
```
Following the current convention in the other modules. We can update the version at the end.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    using AssociatedLinkedListSetLib for AssociatedLinkedListSet;
+
+    struct ERC20SpendLimit {
+        address token;
+        uint256 limit;
+    }
+
+    string internal constant _NAME = "ERC20 Token Limit Module";
+    string internal constant _VERSION = "1.0.0";
+    string internal constant _AUTHOR = "Alchemy & ERC-6900 Authors";
```
adamegyed:
```suggestion
```
The `moduleId` function includes this information.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+        (bytes4 selector, bytes memory callData) = _getSelectorAndCalldata(data);
+
+        if (
+            selector == IModularAccount.execute.selector
+                || selector == IModularAccount.executeWithAuthorization.selector
+                || selector == IAccountExecute.executeUserOp.selector
+        ) {
+            // when calling execute or ERC20 functions directly
+            (address token,, bytes memory innerCalldata) = abi.decode(callData, (address, uint256, bytes));
+            if (_tokenList.contains(msg.sender, SetValue.wrap(bytes30(bytes20(token))))) {
```
adamegyed: nit: I think we forgot to export `CastLib` into `modular-account-libs`. In the meantime, could we create an internal function to do the cast `SetValue.wrap(bytes30(bytes20(token))))`?

Something like `function _toSetValue(address token) internal returns (SetValue)`

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+        }
+        return "";
+    }
+
+    /// @inheritdoc IModule
+    function onInstall(bytes calldata data) external override {
+        (uint32 entityId, ERC20SpendLimit[] memory spendLimits) = abi.decode(data, (uint32, ERC20SpendLimit[]));
+
+        for (uint8 i = 0; i < spendLimits.length; i++) {
+            _tokenList.tryAdd(msg.sender, SetValue.wrap(bytes30(bytes20(spendLimits[i].token))));
```
adamegyed: return value of `tryAdd` isn't checked - should we be checking this?

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    AssociatedLinkedListSet internal _tokenList;
+
+    error ExceededTokenLimit();
+    error ExceededNumberOfEntities();
+    error SelectorNotAllowed();
+    error SpendingRequestNotAllowed(bytes4);
+    error ERC20NotAllowed(address);
+
+    function updateLimits(uint32 entityId, address token, uint256 newLimit) external {
+        _tokenList.tryAdd(msg.sender, SetValue.wrap(bytes30(bytes20(token))));
```
adamegyed: Return value of `tryAdd` also not checked

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+            Call[] memory calls = abi.decode(callData, (Call[]));
+            for (uint256 i = 0; i < calls.length; i++) {
+                if (_tokenList.contains(msg.sender, SetValue.wrap(bytes30(bytes20(calls[i].target))))) {
+                    _decrementLimit(entityId, calls[i].target, calls[i].data);
+                } else {
+                    revert ERC20NotAllowed(calls[i].target);
+                }
+            }
+        } else {
+            revert SpendingRequestNotAllowed(selector);
```
adamegyed, line 72: Does this mean all other selectors are blocked? If so, it would make it impossible for a session key to call `.approve(...)` for a dapp contract, then to invoke the dapp contract itself.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+        for (uint8 i = 0; i < spendLimits.length; i++) {
+            _tokenList.tryAdd(msg.sender, SetValue.wrap(bytes30(bytes20(spendLimits[i].token))));
+            limits[entityId][spendLimits[i].token][msg.sender] = spendLimits[i].limit;
+        }
+    }
+
+    /// @inheritdoc IModule
+    function onUninstall(bytes calldata data) external override {
+        (address token, uint32 entityId) = abi.decode(data, (address, uint32));
+        delete limits[entityId][token][msg.sender];
```
adamegyed, line 97: This seems to only clear the limit, not the list of tokens - how should we handle this? Reset & clear all?

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    function _decrementLimit(uint32 entityId, address token, bytes memory innerCalldata) internal {
+        bytes4 selector;
+        uint256 spend;
+        assembly {
+            selector := mload(add(innerCalldata, 32)) // 0:32 is arr len, 32:36 is selector
+            spend := mload(add(innerCalldata, 68)) // 36:68 is recipient, 68:100 is spend
+        }
+        if (selector == IERC20.transfer.selector || selector == IERC20.approve.selector) {
```
adamegyed: Did you mean to use `_isAllowedERC20Function` here?

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+
+    /// @inheritdoc BaseModule
+    function supportsInterface(bytes4 interfaceId) public view override(BaseModule, IERC165) returns (bool) {
+        return interfaceId == type(IExecutionHookModule).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function _decrementLimit(uint32 entityId, address token, bytes memory innerCalldata) internal {
+        bytes4 selector;
+        uint256 spend;
+        assembly {
```
adamegyed:
```suggestion
        assembly ("memory-safe") {
```
To correctly mark this as memory-safe, we also need to check that the length is at least 100 bytes.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+            selector := mload(add(innerCalldata, 32)) // 0:32 is arr len, 32:36 is selector
+            spend := mload(add(innerCalldata, 68)) // 36:68 is recipient, 68:100 is spend
+        }
+        if (selector == IERC20.transfer.selector || selector == IERC20.approve.selector) {
+            uint256 limit = limits[entityId][token][msg.sender];
+            if (spend > limit) {
+                revert ExceededTokenLimit();
+            }
+            // solhint-disable-next-line reentrancy
+            limits[entityId][token][msg.sender] = limit - spend;
```
adamegyed: Can be `unchecked` because we have the `spend > limit` check above.

adamegyed (commented):
Thanks for working on this! Had some comments.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,160 @@
...
+///     - This module only provide hooks associated with validations.
+///     - For any validation function with this hook installed, all ERC20s without limits specified here will be
+/// reverted.
+///     - Only spending request through the following native execution functions are supported:
+/// IModularAccount.execute, IModularAccount.executeWithAuthorization, IAccountExecute.executeUserOp,
+/// IModularAccount.executeBatch. All other spending request will be reverted.
+///     - this module is opinionated on what selectors (transfer and approve only) can be called for token
+/// contracts to guard against weird edge cases like DAI. You wouldn't be able to use uni v2 pairs directly as the
+/// pair contract is also the LP token contract
+
```
jaypaik: Nit: remove newline

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: That is intended to save the check if it was added already.
We could do a zero address check actually, gonna add that.

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: Token list is shared by all entities of user. So we cannot just delete it.

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: Hmm, what do you want to change here?

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: [Same as above](https://github.com/alchemyplatform/modular-account/pull/184/files#r1772310693) on `updateLimit`.

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: > Does this mean all other selectors are blocked?

It blocks all other selectors native or installed on the account.

> If so, it would make it impossible for a session key to call .approve(...) for a dapp contract, then to invoke the dapp contract itself.

If the dapp installs an execution function `approve` on the account, the SK with ERC20 limit wont be able to call in this case.
The reason to block all other account selectors is that you don't know if the installed execFunc is doing batching or individual actions. So the hook cannot meaningfully check. Definitely not the most flexible solution. Open to suggestions.

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: We will just need to check 68, instead of 100 since the length of bytes are not included.

adamegyed: No, we need to check that it is at least 100. When we do `mload(x)`, it reads bytes [x: x +32), inclusive of the former, exclusive of the latter. So when on line 146 we do `mload(add(innerCalldata, 68))`, that is reading bytes [68:100) of calldata, as listed in the comment, and if the length was > 68 but < 100, then it would be reading garbage values.

src/modules/ERC20TokenLimitModule.sol
adamegyed: we can wrap it in an `unchecked` block, since we have a custom error on underflow.

src/modules/ERC20TokenLimitModule.sol
adamegyed: Why is the list not per-entity?

src/modules/ERC20TokenLimitModule.sol
adamegyed: Oh sorry my bad, I mixed this up with a selector on the target contract. If this is an account contract, I think it's ok to allow other selectors, but not totally sure here. The other execution functions can't call `approve` on behalf of the account unless they use some other validation installed for `execute`/`executeBatch`, at which point it would be a different spending limit than the one we're currently checking.

Does that seem to make sense? Or is there a missing case?

src/modules/ERC20TokenLimitModule.sol
adamegyed: AssociatedLinkedListSet already does that check: https://github.com/erc6900/modular-account-libs/blob/develop/src/libraries/AssociatedLinkedListSetLib.sol#L51

But we'd need to pay attention to the return value to use it

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: That would be problematic since the validation-associated hooks wont run when it is a direct call to self.

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: Good question. I believe it was meant to increase visibility of the states and help checking permissions later. 
I think it is actually not necessary at all.

src/modules/ERC20TokenLimitModule.sol
fangting-alchemy: When you check the length of `innerCalldata`, the correct formatted ones will return 68, instead of 100. Because the first 32 (where the length is stored) is not considered in `innerCalldata.length`.
So when we check, the value to check against is 68, rather than 100.

adamegyed: Ah shoot, my bad, this is a `memory` type, not a `calldata` type, which holds the length in a separate stack var. Yep, we're OK just checking up to 68, because this is already copied into memory.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,143 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+import {UserOperationLib} from "@eth-infinitism/account-abstraction/core/UserOperationLib.sol";
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+
+import {SetValue} from "@erc6900/modular-account-libs/libraries/AssociatedLinkedListSetLib.sol";
```
adamegyed: With the refactor, this import and the `_toSetValue` internal function are no longer used.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,143 @@
...
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
```
adamegyed: The EnumerableSet import and the `using EnumerableSet for EnumerableSet.AddressSet;` statement are both unused now and can be removed. I think the linter sees the `using` statement and thinks it is used, so it misses this case.

src/modules/ERC20TokenLimitModule.sol
```diff
@@ -0,0 +1,143 @@
...
+            for (uint256 i = 0; i < calls.length; i++) {
+                _decrementLimit(entityId, calls[i].target, calls[i].data);
+            }
+        } else {
+            revert SpendingRequestNotAllowed(selector);
+        }
+        return "";
+    }
+
+    /// @inheritdoc IModule
```
adamegyed, line 77: Similar to the natspec comment requests for other modules, could you add the encoding format of `data` to natpspec as an `@dev` comment? This should help by making it show up in docs and on explorers.

adamegyed (approved):
Looks good! Had just small comment and unused import nits.

PR #186, opened 2024-09-23
feat: use custom base account

adamegyed:
## Motivation

We want to minimize code size + gas usage.

## Solution

Optimize the account by:
- Omitting `getNonce`, as this is already public the EP, whose address is retrievable via `entryPoint()`.
- Not collecting return data, nor checking the success code, of sending ether to the EntryPoint.
   - From the 4337 team, we can omit this check: https://github.com/eth-infinitism/account-abstraction/blob/v0.7.0/contracts/core/BaseAccount.sol#L110

We previously inherited from the 4337 team's BaseAccount, this replaces that with a custom implementation.

Also renames `_validateSignature` -> `_validateUserOp` to reduce confusion.


.forge-snapshots/ModularAccount_Runtime_AccountCreation.snap
```diff
@@ -1 +1 @@
-188373
```
fangting-alchemy: 22 is > 0. I will take it.

src/account/BaseAccount.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {IAccount} from "@eth-infinitism/account-abstraction/interfaces/IAccount.sol";
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+
+/// @dev An optimized implementation of a base account contract for ERC-4337.
```
fangting-alchemy, line 8: Can we add the changes we made to the doc?

adamegyed: Added some extra documentation comments about how this contract omits the nonce getter, and uses the internal function `_validateUserOp` for the custom account logic, which is different from the 4337 team's `_validateSignature` named function.

src/account/BaseAccount.sol
```diff
@@ -0,0 +1,54 @@
...
+        _requireFromEntryPoint();
+
+        validationData = _validateUserOp(userOp, userOpHash);
+
+        // Pay the prefund if necessary.
+        // todo: storage-warming optimization if nonzero requiredPrefund
+        assembly ("memory-safe") {
+            if missingAccountFunds {
+                // Ignore failure (it's EntryPoint's job to verify, not the account's).
+                pop(call(gas(), caller(), missingAccountFunds, codesize(), 0x00, codesize(), 0x00))
```
Zer0dot, line 32: technically with push0 the cost of a `CODESIZE` is the same, so we could probably just replace these with zeroes-- although it depends on Solidity playing nice!

Zer0dot (approved):

src/account/BaseAccount.sol
adamegyed: We might still use EVM version Paris, so I kept it as `codesize` for now, should be the same cost & effect. If we do use Shanghai or later, then the 0's afterwards will also get cheaper.

PR #187, opened 2024-09-24
feat: use solady proxies

adamegyed:
## Motivation

Optimizing code size and runtime gas.

## Solution

Switch UUPSUpgradeable from the OZ implementation to the solady implementation. This removes the public function `UPGRADE_INTERFACE_VERSION`, but keeps the implementations of `proxiableUUID` and `upgradeToAndCall`.

For ModularAccount specifically, switch from using OZ `ERC1967` proxies to using solady's LibClone to create ERC1967 proxies. This also reduces codesize on the factory, and prevents us from having to compute the proxy bytecode hash in the constructor. SemiModularAccount was already using this library, as it was necessary for appending immutable args.


gas/modular-account/ModularAccount.gas.t.sol
```diff
@@ -28,19 +28,18 @@ contract ModularAccountGasTest is ModularAccountBenchmarkBase("ModularAccount")
         // Also assert that the event emitted by the factory is correct
         Vm.Log[] memory logs = vm.getRecordedLogs();
 
-        assertEq(logs.length, 5);
+        assertEq(logs.length, 4);
         // Logs:
-        // 0: ERC1967Proxy `Upgraded`
```
adamegyed, line 33: For reviewers: The soladay proxy doesn't emit this on creation. The solady UUPSUpgradeable emits it as expected on any subsequent upgrades.

src/factory/AccountFactory.sol
```diff
@@ -47,24 +42,25 @@ contract AccountFactory is Ownable {
      */
     function createAccount(address owner, uint256 salt, uint32 entityId) external returns (ModularAccount) {
```
Zer0dot, line 43: unrelated, but what was the point of including the entityId here again?

Zer0dot (approved):

gas/modular-account/ModularAccount.gas.t.sol
Zer0dot: Shouldn't be a problem right

src/factory/AccountFactory.sol
adamegyed: Not sure tbh... we could probably either remove it, or offer two functions and have one of them pick a default (likely 0).

gas/modular-account/ModularAccount.gas.t.sol
adamegyed: I think it's ok, we don't use this in any of our offchain services.

PR #188, opened 2024-09-24
feat: deferred install gas benchmarks

Zer0dot:
## Motivation

Installing deferred validations will likely be a fairly high volume action. Benchmarking gas use for a simple case deferred validation grants us insights into gas consumption for this pretty common action. 

## Solution

Simply copy and lightly modify the standard test file (`DeferredValidation.t.sol`) internal functions for deferred validation into the base benchmark gas contract (`ModularAccountBenchmarkBase.sol`.

Then, add the gas tests for deferred validation for both the standard MA and the SMA with bytecode.


gas/modular-account/ModularAccountBenchmarkBase.sol
```diff
@@ -43,4 +58,122 @@ abstract contract ModularAccountBenchmarkBase is BenchmarkBase, ModuleSignatureU
         account1 = factory.createSemiModularAccount(owner1, 0);
         signerValidation = FALLBACK_VALIDATION;
     }
+
+    // Internal Helpers
```
adamegyed: Could we move these helper functions:
- `_buildFullDeferredInstallSig`
- `_computeDomainSeparatorNotDeployed`
- `_getSmaReplaySafeHash`
- `_getDeferredInstallSig`
- `_hashStruct` (might want to rename to be ReplaySafeHash specific)

to the shared mixin contract ModuleSignatureUtils in `test/utils/`? This way, we won't need two definitions for them in here and in `DeferredInstall.t.sol`, so just one location to maintain, and it'll be accessible to both the regular and gas tests.

gas/modular-account/ModularAccountBenchmarkBase.sol
```diff
@@ -2,19 +2,31 @@
...
-import {BenchmarkBase} from "..//BenchmarkBase.sol";
+
+import {BenchmarkBase} from "../BenchmarkBase.sol";
 
 abstract contract ModularAccountBenchmarkBase is BenchmarkBase, ModuleSignatureUtils {
+    using ValidationConfigLib for ValidationConfig;
+
+    bytes32 private constant _INSTALL_VALIDATION_TYPEHASH = keccak256(
+        // solhint-disable-next-line max-line-length
+        "InstallValidation(bytes25 validationConfig,bytes4[] selectors,bytes installData,bytes[] hooks,uint256 nonce,uint48 deadline)"
```
adamegyed: Mini-tip: you can continue a string literal across multiple lines by just putting two `"..."` expressions sequentially, like `"abc" "def" == "abcdef"`.
```suggestion
        "InstallValidation(bytes25 validationConfig,bytes4[] selectors,bytes installData,bytes[] hooks,"
        "uint256 nonce,uint48 deadline)"
```

gas/modular-account/ModularAccountBenchmarkBase.sol
```diff
@@ -43,4 +58,122 @@ abstract contract ModularAccountBenchmarkBase is BenchmarkBase, ModuleSignatureU
...
+         * bytes 129-132          : deferred install validation sig length (e.g. 68)
+         * bytes 133-200 (example): install validation sig data (the data passed to the outer validation to
+         * validate the deferred installation)
+         * bytes 201...           : signature data passed to the newly installed deferred validation to validate
+         * the UO
+         */
+        uint8 outerValidationFlags = 3;
+
+        ValidationConfig deferredConfig = ValidationConfigLib.pack({
+            _module: _mockValidation,
```
adamegyed: Is using mockValidation instead of the real SSV going to result in inaccurate gas benchmarks?

adamegyed (commented):
Looks good! Had some comments.

gas/modular-account/ModularAccountBenchmarkBase.sol
Zer0dot: Yeah for sure, I'll have to do a smol bit of refactoring to use the gas version, which uses a bool instead of an env variable for SMA, in the test-- but I think that's entirely fine. Good call, I was wondering if we had a place to share code between the two dirs!

gas/modular-account/ModularAccountBenchmarkBase.sol
Zer0dot: In general you're probably not going to need to install a second SSV, no? If we use the SSV we're already using, we would end up hitting some warmed storage. We could deploy a copy of it and use that though.

My initial thought was that the benchmark should probably focus on the overhead of using deferred validation, so it defeats the purpose if the installed validation costs a lot of gas and clouds our judgement re: overhead-- but thinking about it now, it actually makes sense to use an SSV because that's what we're benchmarking against. 

I'll go ahead and deploy a new SSV and use that, it should give us the clearest picture relative to the cost of installing a validation normally.

gas/modular-account/ModularAccountBenchmarkBase.sol
Zer0dot: I had no idea what the heck

gas/modular-account/ModularAccountBenchmarkBase.sol
Zer0dot: Okay had to do some cursed refactoring with a lot of params, but it looks okay now. Will take another look when I refactor to use another SSV! [98dcb5b](https://github.com/alchemyplatform/modular-account/pull/188/commits/98dcb5b3dfba6a2239f31c16972f708a2ddff430)

Zer0dot: Okay I think I got it to a place where it's good enough for now. I also refactored the standard tests to use a single signer validation. Feel free to give it one more look!

gas/modular-account/ModularAccountBenchmarkBase.sol
Zer0dot: Should be good to go now!

gas/modular-account/ModularAccountBenchmarkBase.sol
```diff
@@ -9,19 +9,22 @@ import {AccountFactory} from "../../src/factory/AccountFactory.sol";
 
 import {FALLBACK_VALIDATION} from "../../src/helpers/Constants.sol";
 import {ModuleEntity, ModuleEntityLib} from "../../src/libraries/ModuleEntityLib.sol";
+
+import {ValidationConfig, ValidationConfigLib} from "../../src/libraries/ValidationConfigLib.sol";
 import {SingleSignerValidationModule} from "../../src/modules/validation/SingleSignerValidationModule.sol";
 
 import {ModuleSignatureUtils} from "../../test/utils/ModuleSignatureUtils.sol";
-import {BenchmarkBase} from "..//BenchmarkBase.sol";
+
```
adamegyed: nit: `forge fmt` likes to randomly insert newlines when sorting imports, could we remove these? The fmt checker passes after removing these lines.

gas/modular-account/SemiModularAccount.gas.t.sol
```diff
@@ -3,10 +3,12 @@ pragma solidity ^0.8.26;
 
 import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
 import {MessageHashUtils} from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
+
```
adamegyed: also here

test/account/DeferredValidation.t.sol
```diff
@@ -2,38 +2,73 @@
...
-
-import {ValidationConfig, ValidationConfigLib} from "../../src/libraries/ValidationConfigLib.sol";
-
-import {MockUserOpValidationModule} from "../mocks/modules/ValidationModuleMocks.sol";
 import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
 import {MessageHashUtils} from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
 
+import {ModuleEntity, ModuleEntityLib} from "../../src/libraries/ModuleEntityLib.sol";
+import {ValidationConfig, ValidationConfigLib} from "../../src/libraries/ValidationConfigLib.sol";
+
```
adamegyed: and here

adamegyed (approved):
Looks good! Just some small style nits

gas/modular-account/ModularAccountBenchmarkBase.sol
Zer0dot: Done! Wonder why the formatter does that..

PR #189, opened 2024-09-24
feat: add webauthn module

howydev:
1. Add webauthn module
2. Rename "SingleSigner" naming -> ECDSA


src/modules/validation/ECDSAValidationModule.sol
```diff
@@ -25,7 +25,7 @@ import {ISingleSignerValidationModule} from "./ISingleSignerValidationModule.sol
 ///
 /// - This validation supports composition that other validation can relay on entities in this validation
 /// to validate partially or fully.
-contract SingleSignerValidationModule is ISingleSignerValidationModule, ReplaySafeWrapper, BaseModule {
+contract ECDSAValidationModule is IECDSAValidationModule, ReplaySafeWrapper, BaseModule {
```
adamegyed, line 27: Mentioned in the sync earlier this week, we should consider incorporating "K1" somewhere in the name, because technically webauthn also uses ECDSA.

src/modules/validation/IWebauthnValidationModule.sol
```diff
@@ -0,0 +1,31 @@
...
+    /// @param entityId The entityId for the account and the signer.
+    /// @param newX X coordinate of the new signer.
+    /// @param newY Y coordinate of the new signer.
+    /// @param oldX X coordinate of the old signer.
+    /// @param oldY Y coordinate of the old signer.
+    event SignerTransferred(
+        address indexed account,
+        uint32 indexed entityId,
+        uint256 indexed newX,
+        uint256 indexed newY,
```
adamegyed, line 18: I wonder if it would make sense to index on an aggregate type like:
```solidity
struct WebauthnSigner {
    uint256 x;
    uint256 y;
}
```

If we include that as an indexed type, then a hash over the struct would be used as the index value, which would reduce the topic count from 4 to 3. We would probably have to include the newX and newY in the data field, to have it retrievable.

Not sure if this is a net gas savings, might be worth testing in the future.

src/modules/validation/WebauthnValidationModule.sol
```diff
@@ -0,0 +1,148 @@
...
+
+import {IModule} from "@erc6900/reference-implementation/interfaces/IModule.sol";
+import {IValidationModule} from "@erc6900/reference-implementation/interfaces/IValidationModule.sol";
+import {BaseModule} from "@erc6900/reference-implementation/modules/BaseModule.sol";
+import {ReplaySafeWrapper} from "@erc6900/reference-implementation/modules/ReplaySafeWrapper.sol";
+
+import {IWebauthnValidationModule} from "./IWebauthnValidationModule.sol";
+
+/// @title Webauthn Validation
+/// @author ERC-6900 Authors
```
adamegyed: We might want "Alchemy" here, and an acknowledgement for the OSS libraries used and their authors below.

src/modules/validation/WebauthnValidationModule.sol
```diff
@@ -0,0 +1,148 @@
...
+/// @notice This validation enables Webauthn (secp256r1 curve) signature validation. It handles installation by
+/// each entity (entityId).
+/// Note: Uninstallation will NOT disable all installed validation entities. None of the functions are installed on
+/// the account. Account states are to be retrieved from this global singleton directly.
+///
+/// - This validation supports ERC-1271. The signature is valid if it is signed by the owner's private key
+/// (if the owner is an EOA) or if it is a valid ERC-1271 signature from the owner (if the owner is a contract).
```
adamegyed: I think this contract doesn't support contract owners

src/modules/validation/WebauthnValidationModule.sol
```diff
@@ -0,0 +1,148 @@
...
+        return _1271_INVALID;
+    }
+
+    // ┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
+    // ┃    Module interface functions    ┃
+    // ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
+
+    /// @inheritdoc IModule
+    function moduleId() external pure returns (string memory) {
+        return "erc6900.webauthn-validation-module.1.0.0";
```
adamegyed:
```suggestion
        return "alchemy.webauthn-validation-module.1.0.0";
```

test/modules/WebauthnValidationModule.t.sol
```diff
@@ -0,0 +1,156 @@
...
+    EntryPoint ep;
+    WebauthnValidationModule module;
+    address payable account;
+    uint32 entityId = 1;
+    // Example key from Coinbase Smart Wallet
+    uint256 passkeyPrivateKey = uint256(0x03d99692017473e2d631945a812607b23269d85721e0f370b8d3e7d29a874fd2);
+    uint256 x =
+        12_673_873_082_346_130_924_691_454_452_779_514_193_164_883_897_088_292_420_374_917_853_190_248_779_330;
+    uint256 y =
+        18_542_991_761_951_108_740_563_055_453_066_386_026_290_576_689_311_603_472_268_584_080_832_751_656_013;
```
adamegyed: If we bump the version of forge-std to [v1.9.3](https://github.com/foundry-rs/forge-std/releases/tag/v1.9.3) (released earlier today), we can use the cheatcode `vm.publicKeyP256` to derive this programmatically. Not urgent, just a potential devex improvement for the future.

test/modules/WebauthnValidationModule.t.sol
```diff
@@ -0,0 +1,156 @@
...
+
+    // EP Constants
+    uint256 _SIG_VALIDATION_PASSED = 0;
+    uint256 _SIG_VALIDATION_FAILED = 1;
+
+    function setUp() external {
+        ep = new EntryPoint();
+        module = new WebauthnValidationModule();
+        account = payable(address(new ERC1967Proxy(address(new ReferenceModularAccount(ep)), "")));
```
adamegyed: Curious why this uses the reference modular account, over MA?

adamegyed (commented):
🔥

src/modules/validation/IWebauthnValidationModule.sol
howydev: It might result in some savings overall, yeah. I like usage in the contract being consistent, so this change would mean we should change function signatures also. We can abstract the type from contract storage so there wouldn't be any additional cost there, just in calldata

I think either approach works, don't yet see a strong reason why one should be done over the other

test/modules/WebauthnValidationModule.t.sol
howydev: TIL, that's very neat

test/modules/WebauthnValidationModule.t.sol
howydev: good catch, will make the change

src/modules/validation/ECDSAValidationModule.sol
howydev: Thoughts on just having it in the comments? If the contract name has to change, are you thinking of something like `256K1ValidationModule`?

adamegyed: Curious what others think, I was thinking maybe just `ECDSAk1`, or if we wanted to be technical, `P256k1`. But also this isn't a blocker for this PR.

howydev: yeah we'll probably do a big naming thing soon, happy to revisit then!

howydev:
> Review potential Missing/Improper Check on the Admin Address issue that is exposed in the [constructor function](https://app.octane.security/projects/p/5b3592e2-4e19-4c0f-996e-117368e7c109/analysis/61/vulnerabilities/d3eb5c25-6f17-4865-bf54-ff0324e0a8e5)

Not a vuln

test/modules/WebauthnValidationModule.t.sol
```diff
@@ -0,0 +1,155 @@
...
+        12_673_873_082_346_130_924_691_454_452_779_514_193_164_883_897_088_292_420_374_917_853_190_248_779_330;
+    uint256 public y =
+        18_542_991_761_951_108_740_563_055_453_066_386_026_290_576_689_311_603_472_268_584_080_832_751_656_013;
+
+    // EP Constants
+    uint256 internal constant _SIG_VALIDATION_PASSED = 0;
+    uint256 internal constant _SIG_VALIDATION_FAILED = 1;
+
+    function setUp() external {
+        ep = new EntryPoint();
```
adamegyed: I think this should already be available as `entryPoint` from AccountTestBase

adamegyed (approved):
Looks good, one small nit in the test

PR #192, opened 2024-09-25
feat: enable contract creation and update forge snapshots

linnall:
## Motivation

Enable accounts to deploy contracts using both deterministic (`create2`) and non-deterministic deployment (`create`)

## Solution

Added the methods `performCreate` and `performCreate2` to Modular Account


src/account/ModularAccount.sol
```diff
@@ -161,6 +162,71 @@ contract ModularAccount is
...
+    /// @param initCode The initCode to deploy.
+    /// @return createdAddr The created contract address.
+    ///
+    /// @dev Assembly procedure:
+    ///     1. Load the free memory pointer.
+    ///     2. Get the initCode length.
+    ///     3. Copy the initCode from callata to memory at the free memory pointer.
+    ///     4. Create the contract.
+    ///     5. If creation failed (the address returned is zero), revert with CreateFailed().
+    function performCreate(uint256 value, bytes calldata initCode)
```
adamegyed: Both of these functions should have the `wrapNativeFunction` modifier applied, to perform access control checks and run hooks, if any are needed. They should also be added to the list of function where global validation applies.

adamegyed (commented):
Curious about what the other @alchemyplatform/contractoors think, but I think it'd be useful to combine these two into one function with a boolean or enum to pick between `create` and `create2`. This would reduce the number of external functions that the dispatcher needs to consider, and reduce the surface area for hooks. Also had a callout regarding access control, otherwise looks good!

src/account/ModularAccount.sol
linnall: Is the list of functions where global validation applies the list in the comments [here](https://github.com/alchemyplatform/modular-account/pull/192/files#diff-dceb6f7441dc9a457bec76460f00b07827070293d15f11d315f8ca11ec7799baR118)?

adamegyed: It has a code segment too, which is here: https://github.com/alchemyplatform/modular-account/blob/5b4c7866335c1aa2fa49e77ef65407a8cdc8d42a/src/account/ModularAccount.sol#L814-L818

Basically, 6900 maintains separate access control lists for separate functions, which lets users create different levels of security within the account. But for convenience, most core account functions are grouped into a default pool where any validation with the "global" flag can be used, and this is the code segment where it is checked.

fangting-alchemy: Also, should be added to the knownSelectors.

fangting-alchemy: Here are the [known selectors](https://github.com/alchemyplatform/modular-account/blob/develop/src/libraries/KnownSelectorsLib.sol) to avoid selector collision for installation.

adamegyed (approved):
Looks good, thanks for putting this together!

fangting-alchemy (approved):

PR #194, opened 2024-09-26
fix: Fix ERC20TokenLimitModule and add session key gas benchmarks

adamegyed:
## Motivation

We want to benchmark the installation and usage of session keys, which use various hook types. In the process of implementing this, I also discovered a bug with ERC20TokenLimitModule, which needed to be fixed before the benchmark could be done.

## Solution

Add gas benchmarks for using session keys with ModularAccount and SemiModularAccount. This is split up into "cases", which describe permissions configuration, and this PR adds "Case 1" with an allowlist, time range, and ERC-20 spend limit. Future benchmarks can introduce new session key cases.

In the process of implementing this, I discovered that the ERC20TokenLimitModule wasn't able to distinguish between addresses that are tokens with spending limits set to zero, and addresses that are not ERC-20 tokens with spending limits. This PR fixes that by introducing an additional state variable `hasLimit` to the module's state, to distinguish between zero-limit tokens and addresses that are either not tokens, or don't have a spend limit applied.

Also moves ERC20TokenLimitModule to the `permissions/` folder.

The test cases include user op and runtime setups for installing the session key, using it to interact with an arbitrary contract (in this example, a counter), and with the token that has a limit applied.

Right now, there is some code duplication across the MA and SMA benchmarks that may be merged in the future, but it is not yet straightforward.


src/modules/permissions/ERC20TokenLimitModule.sol
```diff
@@ -32,25 +32,19 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
         uint256 limit;
     }
 
-    mapping(uint32 entityId => mapping(address token => mapping(address account => uint256 limit))) public limits;
+    struct SpendLimit {
+        bool hasLimit;
```
fangting-alchemy, line 37: What is the the purpose of `hasLimit`?

`limit` 0 should be able to indicate the same.

src/modules/permissions/ERC20TokenLimitModule.sol
```diff
@@ -107,12 +98,29 @@ contract ERC20TokenLimitModule is BaseModule, IExecutionHookModule {
...
     function supportsInterface(bytes4 interfaceId) public view override(BaseModule, IERC165) returns (bool) {
         return interfaceId == type(IExecutionHookModule).interfaceId || super.supportsInterface(interfaceId);
     }
 
-    function _decrementLimit(uint32 entityId, address token, bytes memory innerCalldata) internal {
+    function _decrementLimitIfApplies(uint32 entityId, address token, bytes memory innerCalldata) internal {
+        SpendLimit storage spendLimit = limits[entityId][token][msg.sender];
+
+        if (!spendLimit.hasLimit) {
+            return;
```
fangting-alchemy, line 122: This is gonna let all tokens not set here pass, which is against the product spec.

>not-limited token by default  not-allowed.

src/modules/permissions/ERC20TokenLimitModule.sol
adamegyed: I described this in the PR description, please read it carefully. The previous setup is not sufficient, in the sample case 1, there are two addresses on the allowlist: the mock ERC-20, and the counter contract. Without the distinction between limit 0 and a non-metered token, it is impossible for the session key to call `increment()` on the counter, because the ERC20LimitModule will incorrectly interpret the Counter contract as a token, and revert because `increment()` is not an allowed ERC-20 function and the calldata is too short. This makes it impossible to use the ERC-20 spending limit with a session key that needs the ability to call any other function besides `.transfer` or `.approve`, including on non-ERC-20 contracts.

In the test cases, this is the line where it would fail with the old implementation of ERC20TokenLimitModule: https://github.com/alchemyplatform/modular-account/blob/e1e15b56fc235527b32f39cc86e03c1b8c21b857/gas/modular-account/ModularAccount.gas.t.sol#L248

src/modules/permissions/ERC20TokenLimitModule.sol
adamegyed: That is an impossible product spec to satisfy, it would make the token spending limit permission incompatible with any contract that is not an ERC-20.

fangting-alchemy: This is exactly why we have [the product spec](https://www.notion.so/alchemotion/Alchemy-MA-V2-648fb97620f44515a18ec5896e3206ef?pvs=4#9a207a675bc24f9792358864d30fb401) and discussed across teams before V2. Seems like we missed this one. 
Let's start a thread to start a discussion to redefine the product feature.

src/modules/permissions/ERC20TokenLimitModule.sol
fangting-alchemy: Thanks for clarify again here. Detailed description is great. When we are in a time crunch, might be nice to includes some bullet points to help the reviewers to read. 

Re content, See comments below.

src/modules/permissions/ERC20TokenLimitModule.sol
```diff
@@ -10,7 +10,7 @@ import {IExecutionHookModule} from "@erc6900/reference-implementation/interfaces
 import {Call, IModularAccount} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
 import {IModule} from "@erc6900/reference-implementation/interfaces/IModule.sol";
 
-import {BaseModule, IERC165} from "./BaseModule.sol";
+import {BaseModule, IERC165} from "../BaseModule.sol";
 
 /// @title ERC20 Token Limit Module
 /// @author Alchemy & ERC-6900 Authors
```
fangting-alchemy: Can you update the docs with the new behavior?

adamegyed: Added!

fangting-alchemy (approved):

PR #195, opened 2024-09-26
feat: add paymaster guard module

fangting-alchemy:
## Motivation

Users have asks features where a validation is only validation is a certain paymaster is used for the transaction.

## Solution

Create a paymaster guard module to provide pre-validation hooks to check a certain paymaster is used.

### DIFF
--- a/src/modules/PaymasterGuardModule.sol
+++ b/src/modules/PaymasterGuardModule.sol
@@ -16,7 +16,7 @@ import {BaseModule, IERC165} from "./BaseModule.sol";
 contract PaymasterGuardModule is BaseModule, IValidationHookModule {
     mapping(uint32 entityId => mapping(address account => address paymaster)) public payamsters;
 
-    error NotAuthorized();
+    error BadPaymasterSpecified();
 
     /// @inheritdoc IModule
     /// @param data should be encoded with the entityId of the validation and the paymaster address that guards the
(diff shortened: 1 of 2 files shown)


src/modules/PaymasterGuardModule.sol
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    /// @inheritdoc IValidationHookModule
+    function preRuntimeValidationHook(uint32, address, uint256, bytes calldata, bytes calldata)
+        external
+        view
+        override
+    {
+        revert NotImplemented();
```
adamegyed: I think we should let this pass, instead of reverting. If it's in the runtime path, it implies some other entity is paying for the gas already.

src/modules/PaymasterGuardModule.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /// @inheritdoc IValidationHookModule
+    function preUserOpValidationHook(uint32 entityId, PackedUserOperation calldata userOp, bytes32)
+        external
+        view
+        override
+        returns (uint256)
+    {
+        address payingPaymaster = address(bytes20(userOp.paymasterAndData));
```
adamegyed: If `userOp.paymasterAndData` is length 0, this will be loaded as `address(0)`, and the hook will pass if the user is not using a paymaster. Do we want to keep that behavior? If so, we should probably document that the magic value of `address(0)` in this module means that the module will enforce that the validation doesn't use a paymaster. Or, if we don't want that behavior, we should load the data as
```solidity
address payingPaymaster = address(bytes20(userOp.paymasterAndData[:20]));
```
which will revert if length < 20.

src/modules/PaymasterGuardModule.sol
```diff
@@ -0,0 +1,78 @@
...
+        if (payingPaymaster == payamsters[entityId][msg.sender]) {
+            return 0;
+        } else {
+            revert NotAuthorized();
```
adamegyed: We might want to use a specific error message for invalid paymaster, instead of the generic `NotAuthorized`, to make it easier to debug.

adamegyed (approved):
Looks good! Had a few small comments, approving to speed up

src/modules/PaymasterGuardModule.sol
fangting-alchemy: Good point.

PR #196, opened 2024-09-26
feat: introduce a storage-only SMA

Zer0dot:
## Motivation
In order to support upgrading to an SMA, we need to implement a variant which only allows the fallback signer to be read from storage, since we have no context about the proxy being used. 

## Solution

Split the SMAs into three contracts:

1. A base contract, with most of the functionality and a single overridable `_getFallbackSignerUnchecked()` function.
2. A storage-only variant which introduces an initializer and only returns the fallback signer in storage in the function override.
3. A storage-and-bytecode variant which returns either the fallback signer in storage if it's nonzero, or reads from bytecode.

The storage-only SMA was not introduced to the factory since it's not an implementation that should ever be used for new accounts as the bytecode variant is always more efficient in those cases.

Lastly, there is an experimental change to refactor the MA itself to pull out the initializer so we no longer need to short-circuit and revert by overriding it. I'm not sure about this change because inheritance is a double-edged sword and confusion grows exponentially. There isn't much difference in the gas benchmarks, and code size is less than 100 bytes smaller for each SMA. However, it is technically the optimal thing to do.


Zer0dot:
ModularAccountBase is the same contract as the ModularAccount from before, but without the initializer, and made abstract.

src/account/ModularAccountBase.sol
```diff
@@ -0,0 +1,912 @@
+// SPDX-License-Identifier: GPL-3.0
```
howydev, line 1: notes: file is identical to the old `ModularAccount.sol`, except `initializeWithValidation` is removed from this

howydev:
hmm, something that should be here should be `KnownSelectors` as a native function, but it doesn't because we don't check against `initializeWithValidation`. We should probably do that (but in a separate PR)

src/account/ModularAccountBase.sol
howydev: just saw that @Zer0dot already made this note, closing

howydev:
on a related note, should the other SMA functions like `setFallbackSignerDisabled` etc also be in a `KnownSelector`s library?

src/account/SemiModularAccountStorage.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+import {SemiModularAccountBase} from "./SemiModularAccountBase.sol";
+// import {AccountStorageInitializable} from "./AccountStorageInitializable.sol";
```
fangting-alchemy:
```suggestion

```

src/account/SemiModularAccountBase.sol
```diff
@@ -82,23 +74,16 @@ contract SemiModularAccount is ModularAccount {
...
-    {
-        revert InitializerDisabled();
-    }
-
     /// @inheritdoc IModularAccount
     function accountId() external pure override returns (string memory) {
         return "alchemy.semi-modular-account.0.0.1";
     }
 
     function replaySafeHash(bytes32 hash) public view virtual returns (bytes32) {
```
fangting-alchemy, line 93: Can you add docs to this func and other funcs too?

src/account/SemiModularAccountBytecode.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+import {LibClone} from "solady/utils/LibClone.sol";
+
+import {SemiModularAccountBase} from "./SemiModularAccountBase.sol";
+
+contract SemiModularAccountBytecode is SemiModularAccountBase {
```
fangting-alchemy, line 16: Can you add docs for this SMA?

src/account/SemiModularAccountStorage.sol
```diff
@@ -0,0 +1,29 @@
...
+
+contract SemiModularAccountStorage is SemiModularAccountBase {
```
fangting-alchemy: Same here. Docs. thanks.

src/account/SemiModularAccountStorage.sol
```diff
@@ -0,0 +1,29 @@
...
+    constructor(IEntryPoint anEntryPoint) SemiModularAccountBase(anEntryPoint) {}
+
+    function initialize(address initialSigner) external initializer {
+        _getSemiModularAccountStorage().fallbackSigner = initialSigner;
+
+        // Note that it's technically possible for the fallback signer in storage to be nonzero before
+        // initialization. However, reading it here would add costs in the vast majority of cases.
+        emit FallbackSignerSet(address(0), initialSigner);
```
fangting-alchemy: I am now thinking if we should overwrite it, instead of make it impossible to upgrade. wdyt?
It either indicate an old fallbackSigner exist or there is a collision.

src/account/SemiModularAccountBase.sol
Zer0dot: yes for sure!

src/account/SemiModularAccountStorage.sol
Zer0dot: Not sure I follow exactly-- could you elaborate? We overwrite the signer, but we make it impossible to re-initialize by making this an initializer, If the idea is to upgrade from an MA to an SMA-S, I guess we could batch the `SetFallbackSigner()` function call? 

Otherwise, we could also make the initialzier here use another slot in a third namespaced storage struct, but that may be overkill. Would this address the concern though?

src/account/SemiModularAccountStorage.sol
```diff
@@ -0,0 +1,28 @@
...
+pragma solidity ^0.8.26;
+
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+import {SemiModularAccountBase} from "./SemiModularAccountBase.sol";
+
+contract SemiModularAccountStorage is SemiModularAccountBase {
+    constructor(IEntryPoint anEntryPoint) SemiModularAccountBase(anEntryPoint) {}
+
+    function initialize(address initialSigner) external initializer {
```
howydev: thoughts on also setting _getSemiModularAccountStorage.fallbackSignerDisabled to true here also? i think theres a super edge case of: account has a disabled fallbackSigner in SMABytecode, its upgraded to something else then upgraded back to SMAStorage

Zer0dot: But if we do that, you won't be able to use the fallback signer at all, right?

gas/modular-account/ModularAccountBenchmarkBase.sol
```diff
@@ -39,8 +39,8 @@ abstract contract ModularAccountBenchmarkBase is BenchmarkBase, ModuleSignatureU
         signerValidation = ModuleEntityLib.pack(address(ecdsaValidationModule), 0);
     }
 
-    function _deploySemiModularAccount1() internal {
-        account1 = factory.createSemiModularAccount(owner1, 0);
+    function _deploySemiModularAccountBytecode1() internal {
+        account1 = ModularAccount(payable(factory.createSemiModularAccount(owner1, 0)));
```
adamegyed, line 43: Does it need this casting?

Zer0dot: Yes, cause we return a `SemiModularAccountBytecode` which doesn't inherit `ModularAccount` in its chain, but it inherits `ModularAccountBase`, i.e. it inherits the MA without initializer, but we use the MA during tests, so we have to cast it or refactor to use MABase during testing (which is fine too but will take time.)

src/account/SemiModularAccountStorage.sol
howydev: my bad! i meant set `fallbackSignerDisabled` to false, in the edge case it was set to true

src/account/SemiModularAccountBase.sol
Zer0dot: Done [2cf576a](https://github.com/alchemyplatform/modular-account/pull/196/commits/2cf576acaf5fa8745a3e3030979e53cbe84cd45c)!

src/account/SemiModularAccountBytecode.sol
Zer0dot: Yup! Done in [2cf576a](https://github.com/alchemyplatform/modular-account/pull/196/commits/2cf576acaf5fa8745a3e3030979e53cbe84cd45c)!

src/account/SemiModularAccountStorage.sol
Zer0dot: Done!  [2cf576a](https://github.com/alchemyplatform/modular-account/pull/196/commits/2cf576acaf5fa8745a3e3030979e53cbe84cd45c)

gas/modular-account/ModularAccountBenchmarkBase.sol
adamegyed: Ah makes sense, I didn't consider the inheritance chain changes from this PR.

src/account/SemiModularAccountStorage.sol
Zer0dot: Ohh interesting, I guess it's an extra SLOAD at least, but it's in the same slot so nbd. Without this, we could have entered a bricked state, good catch!

Zer0dot:
Renamed SMA-S to `SemiModularAccountStorageOnly` to 

a) be more clear
b) prevent a collision with the storage struct type.

I also enforced that, upon initialization, the `fallbackSignerDisabled` is set to `false` h/t @howydev cause this would have allowed a bricked state when:

1. You have an SMA.
2. You disable the fallback signer (set it to `true`).
3. You upgrade to another account and somehow remove your other validations.
4. You upgrade back to an SMA

In step 4, the fallback signer would still be disabled if we don't reset it on initialize, leading to an account with no validations.

src/account/SemiModularAccountStorageOnly.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+import {SemiModularAccountBase} from "./SemiModularAccountBase.sol";
+
+/// @title SemiModularAccountStorageOnly
```
fangting-alchemy, line 8: Why not just `SemiModularAccountStorage`?

Zer0dot: put this in a comment above but we already have a struct called `SemiModularAccountStorage` and this is more precise because `SemiModularAccountBytecode` has both storage & bytecode. Small thing, mostly for the struct name lol

fangting-alchemy (approved):

PR #197, opened 2024-09-27
test: add hook ordering test

adamegyed:
## Motivation

As we start working on changes the account storage, we want to make sure hooks are executed in the correct order. Currently, we don't have a test that asserts this order.

## Solution

Add a test that asserts hook ordering is correct for the common call flows:
- user op / runtime / signature validation
- module exec function / account native function
- with or without `executeUserOp`, and validation-associated exec hooks

The expected order is:
- All pre-validation hooks (in forward order)
- The validation function
- All pre-exec hooks associated w/ validation (in forward order)
- All pre-exec hooks associated w/ selector (in forward order)
- The exec function
- All post-exec hooks associated w/ selector (in reverse order)
- All post-exec hooks associated w/ validation (in reverse order)

To implement this, add a module that logs to storage whenever any module function is invoked, then read the log at the end of the test to assert that the ordering is correct. Then, write a test contract that has all of the permutations of the above cases, and assert the call order after each.

Note that this test only checks the ordering after the first install - it doesn't check behavior for adding hooks, nor for removing individual hooks, which we tentatively plan to support.

### Order checking for signature validation

The external view function `isValidSignature`, and the module functions `validateSignature` and `preSignatureValidationHook`, are all declared `view`. This means the account will invoke the module functions with `staticcall`, and this will make it impossible for the module to write to storage normally. While there are ways to "secretly" store values in the warming of slots, I instead opted to have the module use the `Vm.store` cheatcode, invoked in a non-standard way with a `staticcall`, to append to its own array.

### DIFF
--- a/test/account/HookOrdering.t.sol
+++ b/test/account/HookOrdering.t.sol
@@ -1,0 +1,627 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+
+import {
+    ExecutionManifest,
+    ManifestExecutionHook
+} from "@erc6900/reference-implementation/interfaces/IExecutionModule.sol";
+
+import {DIRECT_CALL_VALIDATION_ENTITYID} from "../../src/helpers/Constants.sol";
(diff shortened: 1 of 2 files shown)


fangting-alchemy (approved):
Thanks for writing this.

PR #199, opened 2024-09-30
feat: add native token module

howydev:


src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,138 @@
...
+        }
+        limits[entityId][msg.sender] = limit - value;
+
+        return "";
+    }
+
+    /// @inheritdoc IModule
+    function onInstall(bytes calldata data) external override {
+        (uint32 entityId, uint256 spendLimits) = abi.decode(data, (uint32, uint256));
+        limits[entityId][msg.sender] = spendLimits;
```
adamegyed: Nit: this shouldn't be plural

```suggestion
        (uint32 entityId, uint256 spendLimit) = abi.decode(data, (uint32, uint256));
        limits[entityId][msg.sender] = spendLimit;
```

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,138 @@
...
+/// @dev e.g. ERC20 paymasters that pull funds from the account)
+/// @notice This module supports a single total native token spend limit.
+/// It tracks a total spend limit across UserOperation gas limits and native token transfers.
+/// If a non whitelisted paymaster is used, UO gas would not cause the limit to decrease.
+/// If a whitelisted paymaster is used, gas is still counted towards the limit
+contract NativeTokenLimitModule is BaseModule, IExecutionHookModule, IValidationHookModule {
+    using UserOperationLib for PackedUserOperation;
+
+    mapping(uint256 entityIds => mapping(address account => uint256 limit)) public limits;
+    mapping(address paymaster => mapping(address account => bool allowed)) public specialPaymasters;
```
adamegyed: Not urgent for this PR,  but I wonder if we can simplify this system by merging it with the PaymasterGuardModule.

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,138 @@
...
+
+    /// @inheritdoc IModule
+    function onInstall(bytes calldata data) external override {
+        (uint32 entityId, uint256 spendLimits) = abi.decode(data, (uint32, uint256));
+        limits[entityId][msg.sender] = spendLimits;
+    }
+
+    /// @inheritdoc IModule
+    function onUninstall(bytes calldata data) external override {
+        // This is the highest entityId that's being used by the account
```
adamegyed: Does this comment still apply?

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,138 @@
...
+        override
+        returns (bytes memory)
+    {
+        (bytes4 selector, bytes memory callData) = _getSelectorAndCalldata(data);
+
+        uint256 value;
+        // Get value being sent
+        if (selector == IModularAccount.execute.selector) {
+            (, value) = abi.decode(callData, (address, uint256));
+        } else if (selector == IModularAccount.executeBatch.selector) {
```
adamegyed: We should also filter for `performCreate`/`performCreate2`, because they are also capable of spending value now.

howydev: hmm this is an interesting note. this means that the module is will only be capable of blocking native functions that it knows about, and if theres another ERC6900 account with another set of native functions that can spend value, this module would not work for them

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,146 @@
...
+import {ModularAccountBase} from "../account/ModularAccountBase.sol";
+import {BaseModule, IERC165} from "./BaseModule.sol";
+import {IExecutionHookModule} from "@erc6900/reference-implementation/interfaces/IExecutionHookModule.sol";
+import {Call, IModularAccount} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
+import {IModule} from "@erc6900/reference-implementation/interfaces/IModule.sol";
+import {IValidationHookModule} from "@erc6900/reference-implementation/interfaces/IValidationHookModule.sol";
+
+/// @title Native Token Limit Module
+/// @author Alchemy
+/// @dev This module only tracks spending across the 4 native functions `execute`, `executeBatch`, `performCreate`,
```
fangting-alchemy: What about `executeUserOp`?

fangting-alchemy: and the install funcs?

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,146 @@
...
+/// @dev e.g. ERC20 paymasters that pull funds from the account)
+/// @notice This module supports a single total native token spend limit.
+/// It tracks a total spend limit across UserOperation gas limits and native token transfers.
+/// If a non whitelisted paymaster is used, UO gas would not cause the limit to decrease.
+/// If a whitelisted paymaster is used, gas is still counted towards the limit
+contract NativeTokenLimitModule is BaseModule, IExecutionHookModule, IValidationHookModule {
+    using UserOperationLib for PackedUserOperation;
+
+    mapping(uint256 entityIds => mapping(address account => uint256 limit)) public limits;
+    mapping(address paymaster => mapping(address account => bool allowed)) public specialPaymasters;
```
fangting-alchemy: Make add some comments here.

```
// paymasters allowed to pull funds from accounts
```

src/modules/NativeTokenLimitModule.sol
fangting-alchemy: I think it is gonna be tricky to merge those paymaster mappings.
Here is for paymasters that are allowed to pull funds from accounts.
The other one in PaymasterGuard is for paymasters to pay for a specific validation function (they don't allow funds pulling).

We can still merge the two modules with merging the paymasters.

src/modules/NativeTokenLimitModule.sol
fangting-alchemy: Also, can you add docs on how this module should be used?
Should both hooks (pre-validation and execution hooks) installed?

Looks like `preUserOpValidationHook` is metering gas spend and `preExecutionHook` is metering value spend.

Why not include both in one hook to make installation easier?

If the goal is to make those two features independent, we should have independent limits. (Still should not affect them being on one hook).

howydev: hmm maybe the wording can be improved here..

for UO validation we track all UOs regardless of the function called in the execution phase, since gas spend for a UO is in native tokens

during the execution phase of a UO, or during the runtime path, there are only 4 functions that can transfer native tokens - execute, executeBatch, and the 2 new deployment functions. The documentation has to mention this, because if the module is installed on a 6900 compliant account with new native functions that can also transfer ether, the module would not be metering any spends through those other native functions. I cant think of a better design to cover that edge case right now..

howydev: the modules design is to track all native gas spend, since i dont think it makes sense to separate "UO gas spend" from "native eth transfers"  since you can "convert" a UO gas limit to native tokens via the session key manually submitting a UO with itself as the bundler

Then, there are 3 cases to consider:
1. UO to a function that doesn't transfer ether - the module wants to deduct UO gas from the limit
2. UO to a function that transfers ether - the module wants to deduct UO gas and the transfer from the limit
3. Runtime call (`executeWithAuthorization`) to function that transfers ether - the module wants to deduct the transfer from the limit

I think the best way to do this is via 2 hooks - pre val hook deducts UO gas for all UOs and exec hook deducts transfer spend for all executions regardless of if it came from a UO or via the runtime path

src/modules/NativeTokenLimitModule.sol
howydev: if its fine, will reword as "paymasters given permissions to pull funds from accounts should be added here" to the suffix to clarify a little more, since it could be misunderstood as "adding a paymaster to this list gives them permissions to pull funds from the account"

src/modules/NativeTokenLimitModule.sol
adamegyed: Yes, I think that is correct.

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+/// If a whitelisted paymaster is used, gas is still counted towards the limit
+contract NativeTokenLimitModule is BaseModule, IExecutionHookModule, IValidationHookModule {
+    using UserOperationLib for PackedUserOperation;
+
+    mapping(uint256 entityIds => mapping(address account => uint256 limit)) public limits;
+    // paymasters given permissions to pull funds from accounts should be added here
+    mapping(address paymaster => mapping(address account => bool allowed)) public specialPaymasters;
+
+    error ExceededNativeTokenLimit();
+    error ExceededNumberOfEntities();
```
fangting-alchemy:
```suggestion

```

not used.

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+
+    function updateLimits(uint32 entityId, uint256 newLimit) external {
+        limits[entityId][msg.sender] = newLimit;
+    }
+
+    function updateSpecialPaymaster(address paymaster, bool allowed) external {
```
fangting-alchemy: Can you add docs for all public funcs?

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+
+    /// @inheritdoc IValidationHookModule
+    function preUserOpValidationHook(uint32 entityId, PackedUserOperation calldata userOp, bytes32)
+        external
+        returns (uint256)
+    {
+        // Decrease limit only if no paymaster is used, or if its a special paymaster
+        if (
+            userOp.paymasterAndData.length == 0
+                || specialPaymasters[address(bytes20(userOp.paymasterAndData[:20]))][msg.sender]
```
fangting-alchemy:
```suggestion
        if (specialPaymasters[address(bytes20(userOp.paymasterAndData[:20]))][msg.sender]
```
Length check not necessary.

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+        ) {
+            uint256 vgl = UserOperationLib.unpackVerificationGasLimit(userOp);
+            uint256 cgl = UserOperationLib.unpackCallGasLimit(userOp);
+            uint256 pvgl;
+            uint256 ppogl;
+            if (userOp.paymasterAndData.length > 0) {
```
fangting-alchemy: This is checked in the outer if statement already.

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+                (, pvgl, ppogl) = UserOperationLib.unpackPaymasterStaticFields(userOp.paymasterAndData);
+            }
+            uint256 totalGas = userOp.preVerificationGas + vgl + cgl + pvgl + ppogl;
+            uint256 usage = totalGas * UserOperationLib.unpackMaxFeePerGas(userOp);
+
+            uint256 limit = limits[entityId][msg.sender];
+            if (usage > limit) {
+                revert ExceededNativeTokenLimit();
+            }
+            limits[entityId][msg.sender] = limit - usage;
```
fangting-alchemy: This will have the issue @adamegyed mentioned at the sync.
If execution fail for some reason, it will record usage incorrectly.

I am not sure if there is a fix for this tho. Maybe call it out in the docs for this.

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+        } else if (selector == IModularAccount.executeBatch.selector) {
+            Call[] memory calls = abi.decode(callData, (Call[]));
+            for (uint256 i = 0; i < calls.length; i++) {
+                value += calls[i].value;
+            }
+        } else if (
+            selector == ModularAccountBase.performCreate.selector
+                || selector == ModularAccountBase.performCreate2.selector
+        ) {
+            value = abi.decode(callData, (uint256));
```
fangting-alchemy: Can you add more tests for this case?

Would this work for those two?
```
function performCreate(uint256 value, bytes calldata initCode)
function performCreate2(uint256 value, bytes calldata initCode, bytes32 salt)
```

src/modules/NativeTokenLimitModule.sol
```diff
@@ -0,0 +1,147 @@
...
+/// @dev This module only tracks spending across the 4 native functions `execute`, `executeBatch`, `performCreate`,
+/// and `performCreate2`. If another ERC6900 account introduces a new native function that spends native tokens,
+/// this module is not able to track it.
+/// @dev Paymasters that use funds from the users account should be whitelisted
+/// @dev e.g. ERC20 paymasters that pull funds from the account)
+/// @notice This module supports a single total native token spend limit.
+/// It tracks a total spend limit across UserOperation gas limits and native token transfers.
+/// If a non whitelisted paymaster is used, UO gas would not cause the limit to decrease.
+/// If a whitelisted paymaster is used, gas is still counted towards the limit
+contract NativeTokenLimitModule is BaseModule, IExecutionHookModule, IValidationHookModule {
```
fangting-alchemy: Can we structure the docs a bit to include to help readers/users:
- all the limits and restrictions of this module (there is one more for gas metering in a below comment).
- how this modules should be used.
[Example](https://github.com/alchemyplatform/modular-account/blob/develop/src/modules/permissions/ERC20TokenLimitModule.sol#L17-L27). It takes me a reading the code to understand how to use this module.

src/modules/NativeTokenLimitModule.sol
howydev: good find!

howydev: actually wait, can u explain this part?

src/modules/NativeTokenLimitModule.sol
howydev: I thought a little about this while designing the spend limits stuff - one way off the top of my head to make it accurate would be:
1. Decrease limits during UO validation as we're doing now, but additionally, save the gas values during validation in some queue data structure
2. We'll use an hook during execution (hook must have UO context, since we need to associate the correct UO execution with it's UO validation) to pop the first value from the queue and check if the UO hash in execution matches with the stored value
3. If the UO hashes match, no validation failed, so we do nothing. If it doesn't match, some executions failed, in this case we increase the spend limits from the saved values (but still net negative since UO validation costs some gas) and keep popping from the queue till values match up

This is pretty inefficient as a solution, but I'm not sure if there's something better

adamegyed: I think for tracking "gas spend", what we currently have is correct behavior - even if execution reverts, the gas is still spent. And reverts during the execution phase undo the "native token spend" from calling `execute` with value, so we should be all set.

We could attempt to get more accurate with a setup like you described, but I think we're OK with this for now, for simplicity.

fangting-alchemy: >even if execution reverts, the gas is still spent.

It wont the accurate one depends on where the revert is. But I think it is an estimate to begin with. So there is gonna be some difference that we will just have to accept and call it out.

src/modules/NativeTokenLimitModule.sol
fangting-alchemy: `userOp.paymasterAndData[:20]` will revert if length is < 20. Unless the goal is to not revert, the length check can be skipped.

howydev: Yeah I think we don't want to revert here, so it will be able to handle the cases in which no paymaster is used

src/modules/NativeTokenLimitModule.sol
howydev: from above + based on the implementation of [unpackPaymasterStaticFields](https://github.com/eth-infinitism/account-abstraction/blob/6f02f5a28a20e804d0410b4b5b570dd4b076dcf9/contracts/core/UserOperationLib.sol#L120-L128) we need the length check, else it would revert

src/modules/NativeTokenLimitModule.sol
howydev: if we have to be inaccurate, erring on the side of having the limit be an upper bound instead of a lower bound would be safer

howydev: +1 to calling it out in docs

src/modules/NativeTokenLimitModule.sol
howydev: updated! lmk if this is still unclear

fangting-alchemy (approved):

jaypaik:
Should this module be moved under `modules/permissions/`?

PR #200, opened 2024-09-30
feat: migrate account storage to LLS

adamegyed:
## Motivation

We want to reduce the installation costs of new validation functions and hooks. In the future, we also want to support removing arbitrary hooks, rather than clearing all hooks. 

## Solution

Replace the EnumerableSet storage data structures in the account with LinkedListSet. This replaces `validationData.executionHooks`, `executionData.executionHooks`, and `executionData.selectors`.

Update the installation and usage flow in ModuleManagerInternals and ModularAccount.

To preserve the same execution order as before, we need to invert the execution order of pre/post hooks.

Also adds a testing utility for pretty-printing storage accesses. This isn't currently used in the tests, since it's primarily for debugging.

### Interpreting benchmark results

This change greatly decreases the gas costs of installing session keys (validations with permission hooks), which is expected because it reduces the number of storage slots to initialize (change from zero to nonzero).

This change also slightly decreases the session key usage gas costs (~2k gas), which is likely from one less storage read. My suspicious is that this is due to the `hasNext` flag in LLS, but I'm not totally sure.

This change slightly increases the gas cost of using the account without hooks (~500 gas), which I'm not sure how to interpret. I verified in the tests that the number of storage accesses in the no-hooks case is the same as before this change, so it is possibly from the change in memory management for holding pre/post hooks.. We still have many opportunities to optimize this.


.forge-snapshots/SemiModularAccount_UserOp_NativeTransfer.snap
```diff
@@ -1 +1 @@
-162687
```
fangting-alchemy: is there a way to group those stats? SMA to a file, MA to a file etc.

adamegyed: Not easily, unless we fork and maintain a modified copy of [forge-gas-snapshot](https://github.com/marktoda/forge-gas-snapshot).

src/account/AccountStorage.sol
```diff
@@ -57,44 +61,35 @@ function getAccountStorage() pure returns (AccountStorage storage _storage) {
...
 }
 
-/// @dev Helper function to get all elements of a set into memory.
-function toModuleEntityArray(EnumerableSet.Bytes32Set storage set) view returns (ModuleEntity[] memory) {
-    uint256 length = set.length();
-    ModuleEntity[] memory result = new ModuleEntity[](length);
-    for (uint256 i = 0; i < length; ++i) {
-        bytes32 key = set.at(i);
-        result[i] = ModuleEntity.wrap(bytes24(key));
+function toHookConfigArray(LinkedListSet storage set) view returns (HookConfig[] memory) {
```
fangting-alchemy, line 72: ~~Looks like this is not used at all.~~

src/account/ModularAccount.sol
```diff
@@ -580,32 +580,40 @@ contract ModularAccount is
         _execRuntimeValidation(runtimeValidationFunction, callData, authorizationData);
     }
 
-    function _doPreHooks(EnumerableSet.Bytes32Set storage executionHooks, bytes memory data)
+    function _doPreHooks(LinkedListSet storage executionHooks, bytes memory data)
         internal
         returns (PostExecToRun[] memory postHooksToRun)
     {
-        uint256 hooksLength = executionHooks.length();
+        HookConfig[] memory hooks = toHookConfigArray(executionHooks);
```
fangting-alchemy, line 587: This probably will cost slight more gas (at least the loop overhead) compare to how it was done before. The one loop with n ops now becomes 2 loops (one n-1ops, one 1 ops).

src/account/ModularAccount.sol
```diff
@@ -462,7 +462,7 @@ contract ModularAccount is
         // `executeUserOp`. This check must be here because if context isn't passed, we can't tell in execution
         // which hooks should have ran.
         if (
-            getAccountStorage().validationData[validationFunction].executionHooks.length() > 0
+            !getAccountStorage().validationData[validationFunction].executionHooks.isEmpty()
```
fangting-alchemy: This probably cost more than length check looking at [what it is doing](https://github.com/erc6900/modular-account-libs/blob/develop/src/libraries/LinkedListSetLib.sol#L206-L209).

fangting-alchemy: weird lots of runtime calls also cost more gas.

adamegyed: > This probably cost more than length check

We could measure it, but it should be pretty similar - they're both 1 keccak and 1 sload

fangting-alchemy: 388 for UO path, 498 for runtime path.

src/account/AccountStorage.sol
adamegyed: This comment did help me realize that a few other functions that were unused in this file - removing them now

src/account/ModularAccount.sol
adamegyed: Yeah I see the calls are more expensive. Looking at it now, I think the reason might be that we were technically doing something wrong in the before case - we were loading pre hooks from storage before running each, which means a pre-hook could modify state and cause a different pre-hook to run afterwards. This PR fixes that by loading the pre-hooks into memory before starting any execution, which fixes the issue but adds gas costs.

In a future optimization, I think we can reuse the `HookConfig[] memory` between pre hooks and post hooks, to prevent needing that extra copying loop before running the pre hooks.

adamegyed:
I rebased to `develop`, so now some changes are moved over to ModularAccountBase, which may hide some comments from the GitHub viewer.

src/account/ModularAccountBase.sol
```diff
@@ -638,32 +638,40 @@ abstract contract ModularAccountBase is
...
+                ++i;
+            }
         }
 
         // Run the pre hooks and copy their return data to the post hooks array, if an associated post exec hook
         // exists.
-        for (uint256 i = 0; i < hooksLength; ++i) {
-            HookConfig hookConfig = toHookConfig(executionHooks.at(i));
+        for (uint256 i = hooksLength; i > 0;) {
+            // Decrement here, instead of in the loop body, to handle the case where length is 0.
```
Zer0dot: I know we do the same thing somewhere else-- but maybe we can change this comment to something like this?
```suggestion
            // Decrement here, instead of in the loop body, to convert from length to an index.
```

src/account/ModularAccountView.sol
```diff
@@ -40,10 +42,13 @@ abstract contract ModularAccountView is IModularAccountView {
...
-            uint256 executionHooksLen = executionData.executionHooks.length();
-            data.executionHooks = new HookConfig[](executionHooksLen);
-            for (uint256 i = 0; i < executionHooksLen; ++i) {
-                data.executionHooks[i] = toHookConfig(executionData.executionHooks.at(i));
+            // Todo: optimize this array reverse.
+            SetValue[] memory hooks = executionData.executionHooks.getAll();
+            uint256 hooksLength = hooks.length;
+            data.executionHooks = new HookConfig[](hooksLength);
+
+            for (uint256 i = 0; i < hooksLength; ++i) {
```
Zer0dot, line 50:
```suggestion
            for (uint256 i = 1; i < hooksLength + 1; ++i) {
```
Maybe this? Then remove the -1? (and maybe cache hooksLength + 1 in my example cause you know how solc is sometimes)

Also, why does it need to be reversed-- if it didn't, we could use the new `toHookConfigArray()` function yeah?

Zer0dot: Same for the other uses

src/account/ModuleManagerInternals.sol
```diff
@@ -108,14 +108,15 @@ abstract contract ModuleManagerInternals is IModularAccount {
...
+    function _addExecHooks(LinkedListSet storage hooks, HookConfig hookConfig) internal {
+        if (!hooks.tryAdd(toSetValue(hookConfig))) {
             revert ExecutionHookAlreadySet(hookConfig);
         }
     }
 
-    function _removeExecHooks(EnumerableSet.Bytes32Set storage hooks, HookConfig hookConfig) internal {
-        hooks.remove(toSetValue(hookConfig));
+    function _removeExecHooks(LinkedListSet storage hooks, HookConfig hookConfig) internal {
+        // Todo: use predecessor
```
Zer0dot, line 118: How do you think we should pass the predecessor here?  I think the exec hooks to be removed are passed via the manifest, so are we going to change the manifest?

 Or is the idea to introduce a new top-level function which lets users pass a predecessor in calldata?

test/utils/StorageAccesses.sol
```diff
@@ -0,0 +1,118 @@
...
+            console.log(string.concat(formatSlot(reads[j], slotNames)));
+        }
+
+        console.log("Storage writes: %s", writes.length);
+        for (uint256 j = 0; j < writes.length; j++) {
+            console.log(string.concat(formatSlot(writes[j], slotNames)));
+        }
+    }
+
+    function formatSlot(bytes32 slot, PredictedSlotNames memory slotNames) internal pure returns (string memory) {
```
Zer0dot, line 51: I'm not sure the naming `formatSlot` makes sense here-- what if we renamed it like so: 
```suggestion
    function getSlotName(bytes32 slot, PredictedSlotNames memory slotNames) internal pure returns (string memory) {
```

src/account/ModularAccountView.sol
adamegyed: If we switch that, we also have to change the `hooks[i]` to `hooks[i - 1]`, so we're not getting rid of the `1` literal. I'd prefer to keep it as-is.

src/account/ModuleManagerInternals.sol
adamegyed: Yeah, at best we _may_ get a predecessor, for the `uninstallValidation` case when removing a hook. I think we will need to keep a code path for the "no predecessor" case.

test/utils/StorageAccesses.sol
adamegyed: It doesn't always get a slot name, because there may not be one. I called it "format" to indicate "string formatting", is there another name you think fits better?

src/account/ModularAccountView.sol
Zer0dot: oh oops, completely slipped my mind lol

test/utils/StorageAccesses.sol
Zer0dot: fine as is then-- can't come up with a better name without it being longer like "getSlotNameIfExists" or something lol

Zer0dot (approved):

PR #203, opened 2024-09-30
feat: skip return data collection on self-call

adamegyed:
## Motivation

We want to optimize the `executeUserOp` call path. Solidity does some undesirable behavior in its "low level calls" of the form `address(target).call(...)`, where it always allocates memory for return data and copies it over, even if it goes completely unused. The account never needs to pass return data to the EntryPoint.

Additionally, the contract organization of AccountExecutor doesn't make much sense, as it only contains one function `_exec`. 

## Solution

Change `executeUserOp` from doing `address(this).call` to performing the call in inline assembly, skiping the return data collection unless there is a revert, at which point the call is bubbled up.

Also, reorganize AccountExecutor, this new self-call, and the free function`collectReturnData` into a new library calle `ExecutionLib`.

Note that we can't (yet) safely apply this same optimization to `executeWithAuthorization`, because it may be called by an entity that cares about the resulting data.


src/libraries/ExecutionLib.sol
```diff
@@ -0,0 +1,72 @@
...
+
+library ExecutionLib {
+    /// @param target The address of the contract to call.
+    /// @param value The value to send with the call.
+    /// @param data The call data.
+    /// @return result The return data of the call, or the error message from the call if call reverts.
+    function exec(address target, uint256 value, bytes memory data) internal returns (bytes memory result) {
+        // Manually call, collecting return data.
+        assembly ("memory-safe") {
+            let success := call(gas(), target, value, add(data, 0x20), mload(data), codesize(), 0)
```
Zer0dot, line 12: qq-- why codesize()?

adamegyed: This follows the calling convention from solady. On EVM versions Paris and before, it's cheaper and less codesize than `push1 0x00`, and because the copied length is 0, it doesn't affect memory expansion. Alternatively, we could use `returndatasize`, `caller`, `origin`, or any other 2-gas opcode, but solady's convention of using `codesize` makes it more obvious that this is just a cheap random value.

For post-Paris EVM versions, the zeros also get replaced with `push0`, so they also become equivalent in cost.

Zer0dot: That's what I was thinking-- it was familiar to me, now I remember it was from Solady. This makes sense. Happy to hear this value doesn't matter as long as the size is zero.

Zer0dot (approved):
This is a great change!

PR #204, opened 2024-09-30
feat: direct call validation for sma

Zer0dot:
## Motivation

We'd like to support the direct-call validation path for SMA fallback signers.

## Solution

Add logic to return true if the validation used is the fallback signer with the maximum entity Id (direct call entity ID).

Some considerations:

- Do we need the msg.sender check?
- Is there a more efficient way to hook this into the logic, instead of always checking in `_isValidationGlobal()`?
- How does this interact if we try to add hooks or install/uninstall the validation equal to the fallback signer with the maximum entity ID?


src/account/SemiModularAccountBase.sol
```diff
@@ -157,7 +157,28 @@ abstract contract SemiModularAccountBase is ModularAccountBase {
...
+        // Retrieve the fallback signer.
+        address fallbackSigner = _retrieveFallbackSignerUnchecked(smaStorage);
+
+        // Compute the direct call validation key.
+        ModuleEntity fallbackDirectCallValidation =
+            ModuleEntityLib.pack(fallbackSigner, DIRECT_CALL_VALIDATION_ENTITYID);
+
+        // Return true if the validation function passed is the fallback direct call validation key, and the sender
+        // is the fallback signer. This enforces that context is a
+        return validationFunction.eq(fallbackDirectCallValidation) && msg.sender == fallbackSigner;
```
adamegyed, line 181: Do we need to `validationFunction.eq(fallbackDirectCallValidation)` check? I'm trying to figure out where this case is reached, and I think it will happen from here in ModularAccountBase:

https://github.com/alchemyplatform/modular-account/blob/1aeaf28350f808fd4e413b21cbedfbf9f63d089e/src/account/ModularAccountBase.sol#L762-L765

At which point, we've already pulled in the `msg.sender`.

Could this check be replaced with:
```solidity
(address validationModule, uint32 entityId) = validationFunction.unpack();
return validationModule == fallbackSigner && entityId == DIRECT_CALL_VALIDATION_ENTITYID;
```
?

Does that make sense to do?

Zer0dot: You're right, but it's also possible to pass this computed direct call validation in another code path, which would rely on the actual function call for validation failing-- which isn't a guarantee. Is that right?

adamegyed: Could you expand on that - in which case would the validation call not fail, when it should?

I think the function `_isValidationGlobal` is only checking for validation applicability, not for the execution of the validaiton logic itself, which happens in `_execUserOpValidation`, `_execRuntimeValidation`, and `_exec1271Validation`. The only case where the `_exec*` step is skipped is the direct call validation, which only checks applicability.

Zer0dot: Right, but isn't it possible for a user to pass this validation (fallbackSigner+special entityID) to a runtime validation flow, which would pass as applicable, then if the fallback signer implements logic or has some sort of fallback logic (and we skip the interface checks/etc) it could potentially pass, or am I completely off?

Zer0dot: Also, isn't 
```
(address validationModule, uint32 entityId) = validationFunction.unpack();
return validationModule == fallbackSigner && entityId == DIRECT_CALL_VALIDATION_ENTITYID;
```

Equivalent to the .eq check? Are you referring to the msg.sender check?

adamegyed: > a user to pass this validation (fallbackSigner+special entityID) to a runtime validation flow, which would pass as applicable, then if the fallback signer implements logic or has some sort of fallback logic (and we skip the interface checks/etc) it could potentially pass

Ah yes... this is a potential risk if the fallback signer implements the `IValidationModule` functions. So this makes sense - you don't change the validation to a new one, you keep using the fallbackSigner magic value for validation, there's just a different check that reinterprets it based on `msg.sender`.

> isn't [...] equivalent to the .eq check?

Yeah I think it structurally is, it just skips the extra `msg.sender` check. But I now see that you're covering a different case, so it should be good.

adamegyed (approved):
Looks good.

src/account/SemiModularAccountBase.sol
Zer0dot: sweet, thanks!

PR #205, opened 2024-09-30
refactor: clean up modules to be consistent

fangting-alchemy:
## Motivation

Modules structure and formats are not consistent across the code base.

## Solution

Refactor to be consistent:
* move permissions modules to permissions.
* remove interfaces of modules.


adamegyed (approved):
Looks good, thanks for cleaning this up and making it consistent!

PR #207, opened 2024-10-01
fix: fix SMA test and CI

adamegyed:
## Motivation

The script `pnpm prep` is no longer working, due to an SMA test failure. This accidentally made its way to `develop` because it was missing from CI.

## Solution

Fix the counterfactual address generation in the tests for `performCreate`/`performCreate2`.

Add SMA tests to the CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,6 +24,9 @@ jobs:
       - name: Run tests
         run: FOUNDRY_PROFILE=optimized-test-deep forge test -vvv
 
+      - name: Run SMA tests
+        run: FOUNDRY_PROFILE=optimized-test-deep SMA_TEST=true forge test -vvv
+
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -38,4 +41,7 @@ jobs:
         run: forge build
 
       - name: Run tests
-        run: forge test -vvv
\ No newline at end of file
+        run: forge test -vvv
+
+      - name: Run tests
```
Zer0dot: Maybe we should rename this?
```suggestion
      - name: Run SMA tests
```

Zer0dot (approved):

.github/workflows/test.yml
adamegyed: Oops yes, thank you!

PR #209, opened 2024-10-01
feat: add token receiver into MABase

fangting-alchemy:
## Motivation

We have token receiver callback methods in a module right now. It is not part of the initialization of the accounts.

MA users will need to install it to use. However, Most users will want this feature out of the box. This makes it inconvenient and cost gas for users to have to actively enable.

## Solution

Move this widely used feature onto the account.

Also add test for the new funcs.

**Note**: weird gas cost increased. None of the code change affects the code paths tested there. Did some experiments. The gas cost increase simply because of the inclusion of `TokenReceiver` into base.


test/account/TokenReceiver.t.sol
```diff
@@ -0,0 +1,67 @@
...
+
+import {MockERC1155} from "../mocks/MockERC1155.sol";
+import {MockERC721} from "../mocks/MockERC721.sol";
+import {AccountTestBase} from "../utils/AccountTestBase.sol";
+import {TEST_DEFAULT_VALIDATION_ENTITY_ID} from "../utils/TestConstants.sol";
+
+contract TokenReceiverTest is AccountTestBase {
+    address public owner;
+    uint256 public ownerKey;
+    ModularAccount public account;
```
adamegyed: AccountTestBase already has `account1`, `owner1`, `owner1Key`, so we don't need to re-declare them here.

adamegyed (approved):
Looks good! 1 small nit on the test setup.

PR #210, opened 2024-10-01
feat: bump reference-implementation version and fix errors

jaypaik:
Relevant updates:
- Bump dependency versions:
  - `"@erc6900/modular-account-libs": "github:erc6900/modular-account-libs#v0.8.0-rc.1"`
  - `"@erc6900/reference-implementation": "github:erc6900/reference-implementation#v0.8.0-rc.3"`
- `executeWithAuthorization` -> `executeWithRuntimeValidation`
- AccountStorage updated to use `HookConfig[] validationHooks` instead of `ModuleEntity[] preValidationHooks`
  - Question here: should we just keep storage to `ModuleEntity[]` and do the transformation to `HookConfig[]` in the view function instead?


test/mocks/modules/ReturnDataModuleMocks.sol
```diff
@@ -101,9 +101,9 @@ contract ResultConsumerModule is IExecutionModule, BaseModule, IValidationModule
     }
 
     // Check the return data through the execute with authorization case
-    function checkResultExecuteWithAuthorization(address target, bytes32 expected) external returns (bool) {
+    function checkResultexecuteWithRuntimeValidation(address target, bytes32 expected) external returns (bool) {
```
adamegyed:
```suggestion
    function checkResultExecuteWithRuntimeValidation(address target, bytes32 expected) external returns (bool) {
```

test/mocks/modules/ReturnDataModuleMocks.sol
```diff
@@ -129,7 +129,7 @@ contract ResultConsumerModule is IExecutionModule, BaseModule, IValidationModule
             allowGlobalValidation: false
         });
         manifest.executionFunctions[1] = ManifestExecutionFunction({
-            executionSelector: this.checkResultExecuteWithAuthorization.selector,
+            executionSelector: this.checkResultexecuteWithRuntimeValidation.selector,
```
adamegyed:
```suggestion
            executionSelector: this.checkResultExecuteWithRuntimeValidation.selector,
```

adamegyed (approved):
Looks good, 1 small capitalization nit.

> should we just keep storage to ModuleEntity[] and do the transformation to HookConfig[] in the view function instead?

I'm OK with either, but I have a slight suspicion that the version with`HookConfig` is what's causing the marginal increase in gas costs. We can change it if we find it improves performance.

test/mocks/modules/ReturnDataModuleMocks.sol
jaypaik: Great catches!

jaypaik:
> I'm OK with either, but I have a slight suspicion that the version withHookConfig is what's causing the marginal increase in gas costs. We can change it if we find it improves performance.

Tried this and actually seems cheaper on average to store HookConfig. Reverting back.

PR #211, opened 2024-10-01
feat: change fallback validation magic value to zero

Zer0dot:
## Motivation

Currently, the fallback validation is equivalent to the maximum value, which means calldata costs are maximized for these 24 bytes. We should change it to, instead, use zero bytes and save additional calldata costs.

## Solution

Update the fallback validation magic value to be a wrapper of `bytes24(0)`. This has repercussions because:

1. Solidity inherently treats uninitialized values as zero.
2. We used `address(0)` as a generic invalid EOA address throughout our tests. This didn't work with our current testing setup because we need to have both the fallback validation and the ECDSAValidation at the same address for testing. In order to fix this, I added a constant `CODELESS_ADDRESS` nonzero address value to be used instead.

When reviewing, please consider implications of using this as a relevant value and if there are any additional repercussions. 

An alternative would be to hardcode a simple one-byte value for both the module address and the entity ID. This would increase calldata costs by ~24 gas, but potentially avoid certain issues.


test/modules/ERC20TokenLimitModule.t.sol
```diff
@@ -18,9 +18,10 @@ import {ERC20TokenLimitModule} from "../../src/modules/permissions/ERC20TokenLim
 
 import {MockModule} from "../mocks/modules/MockModule.sol";
 import {AccountTestBase} from "../utils/AccountTestBase.sol";
+import {CODELESS_ADDRESS} from "../utils/TestConstants.sol";
 
 contract ERC20TokenLimitModuleTest is AccountTestBase {
-    address public recipient = address(1);
```
fangting-alchemy, line 23: Why do we need to change this guy?

src/modules/permissions/PaymasterGuardModule.sol
```diff
@@ -14,7 +14,7 @@ import {BaseModule, IERC165} from "../BaseModule.sol";
 /// used.
 /// - If this hook is installed, and no paymaster is setup, all requests will revert
 contract PaymasterGuardModule is BaseModule, IValidationHookModule {
-    mapping(uint32 entityId => mapping(address account => address paymaster)) public payamsters;
+    mapping(uint32 entityId => mapping(address account => address paymaster)) public paymasters;
```
fangting-alchemy, line 17: Thanks for fixing the typo.

fangting-alchemy:
It looks good. Worth the savings.

test/modules/ERC20TokenLimitModule.t.sol
Zer0dot: Ah this was actually a legacy change because I had etched code at address(0) lol

src/modules/permissions/PaymasterGuardModule.sol
Zer0dot: np!

adamegyed (approved):
Looks good, thanks for doing this!

Could you also update the `test` script in `package.json` to not do the SMA test?

PR #212, opened 2024-10-02
fix: fix AllowlistModule with selector wildcard case and update naming

fangting-alchemy:
## Motivation

Our current `AllowlistModule` does not support selector wildcard (all address allowed).

## Solution

- Fix `AllowlistModule` to allow selector wildcard (all address allowed). It now supports all three cases:
   - specific addresss + specific selectors
   - specific addresss + any selectors
   - any addresss + specific selectors 
- Added new public functions for updating and deleting allowlist
- Renamed some vars and funcs

Note: certain use cases incur extra gas cost, which is expected as it adds a storage read. Open to suggestions if you have a better way to do this.


src/modules/permissions/AllowlistModule.sol
```diff
@@ -172,21 +211,25 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
         internal
         view
     {
-        AllowlistEntry storage entry = targetAllowlist[entityId][target][account];
+        if (data.length < 4) {
+            revert NoSelectorSpecified();
+        }
```
adamegyed: We shouldn't always enforce this - this would make it impossible to do a "native token transfer only" transaction while using the allowlist.

src/modules/permissions/AllowlistModule.sol
```diff
@@ -172,21 +211,25 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
...
+
+        bytes4 selector = bytes4(data);
+        if (selectorAllowlist[entityId][selector][address(0)][account]) {
+            // selector wildcard case, any address is allowed
```
adamegyed: I thought the intended design was to still have an allowlist for this case, but to allow selectors from that list. This setting allows the validation function to bypass the address allowlist.

adamegyed (commented):
The update functions look good, had some questions about the checking logic

src/modules/permissions/AllowlistModule.sol
fangting-alchemy: That is the intended design. `address(0)` only exist in this data path if addresses is ignored.

Checkout the [docs above](https://github.com/alchemyplatform/modular-account/pull/212/files#diff-b651f578eb5421786bc3c9eecaae17d80c38d55865d1b17ff56812d45d39a459R23) for feature details.

src/modules/permissions/AllowlistModule.sol
```diff
@@ -98,16 +102,48 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
...
+    /// be overwritten.
+    /// @param entityId The entity ID to initialize the allowlist for.
+    /// @param inputs The allowlist inputs data to update.
+    function updateAllowlist(uint32 entityId, AllowlistInput[] memory inputs) public {
+        for (uint256 i = 0; i < inputs.length; i++) {
+            AllowlistInput memory input = inputs[i];
+            if (input.target == address(0)) {
+                // wildcard case for selectors, any address can be called for the selector
+                for (uint256 j = 0; j < input.selectors.length; j++) {
+                    setSelectorAllowlist(entityId, input.target, input.selectors[j], true);
```
Zer0dot:
```suggestion
                    setSelectorAllowlist(entityId, address(0), input.selectors[j], true);
```

My hunch is that this is slightly more efficient, but I'm not actually sure.

fangting-alchemy: Yup, 3 gas.

src/modules/permissions/AllowlistModule.sol
```diff
@@ -172,21 +211,26 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
...
+            if (selectorAllowlist[entityId][selector][address(0)][account]) {
+                // selector wildcard case, any address is allowed
+                return;
+            }
+
+            revert AddressNotAllowed();
         }
 
         if (hasSelectorAllowlist) {
             if (data.length < 4) {
```
Zer0dot, line 229: Does it make sense to keep this check here if we cast data to a selector earlier? Maybe we can move it up, wdyt?

src/modules/permissions/AllowlistModule.sol
```diff
@@ -172,21 +211,26 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
         internal
         view
     {
-        AllowlistEntry storage entry = targetAllowlist[entityId][target][account];
+        bytes4 selector = bytes4(data);
```
Zer0dot, line 214: I think we can move this down to the if statement when it's first needed right? This way we save the declaration in case it's an allowed address without a selector list.

src/modules/permissions/AllowlistModule.sol
fangting-alchemy: We need to allow the nativeToken spend case where data length will be 0. See comment [here](https://github.com/alchemyplatform/modular-account/pull/212#discussion_r1787913333).

src/modules/permissions/AllowlistModule.sol
fangting-alchemy: It needs to be in the beginning because it is used in a later if statement. We will have to declare twice in each if we don't want to do it in the beginning.

src/modules/permissions/AllowlistModule.sol
Zer0dot: lol nice

src/modules/permissions/AllowlistModule.sol
Zer0dot: Ohhh makes sense!

src/modules/permissions/AllowlistModule.sol
Zer0dot: Oh oops right, sounds good!

src/modules/permissions/AllowlistModule.sol
```diff
@@ -12,61 +12,65 @@ import {BaseModule} from "../../modules/BaseModule.sol";
...
-/// `IModularAccount.executeBatch` functions.
+/// @notice This module allows for the setting and enforcement of allowlists for validation functions.
+///    - These allowlists can specify which addresses and or selectors can be called by the validation function. It
+/// supports:
+///       - specific addresss + specific selectors
+///       - specific addresss + any selectors
+///       - any addresss + specific selectors
+///    - These restrictions only apply to the `IModularAccount.execute` and `IModularAccount.executeBatch`
+/// functions.
+///    - The order of permission checking:
```
Zer0dot, line 23: Think this needs to be updated @fangting-alchemy

Zer0dot (approved):
lgtm past the minor comment about the comment!

PR #215, opened 2024-10-02
feat: skip signature length encoding on final sig

adamegyed:
## Motivation

In SparseCalldataSegmentLib, we treat the final signature (sent to validation, instead of pre-validation hooks), the same as the per-hook data segments. This includes encoding a length for the segment, and introduces the possibility of a non-canonical encoding, if the actual signature length exceeds the declared length.

## Solution

To save gas on length checking, and on calldata size of signatures, we can rearrange how SparseCalldataSegmentLib decodes segments: switch the order of the index and the length. Then, we can omit the length of the final signature segment – it is just defined to be of the entire remaining length.

This saves on execution + calldata gas costs, and removes one possible non-canonical encoding format.

Also updates some existing tests to better use signature utils.


src/libraries/SparseCalldataSegmentLib.sol
```diff
@@ -26,14 +26,14 @@ library SparseCalldataSegmentLib {
         returns (bytes calldata segment, bytes calldata remainder)
     {
         // The first 4 bytes hold the length of the segment, excluding the index.
-        uint32 length = uint32(bytes4(source[:4]));
+        uint32 length = uint32(bytes4(source[1:5]));
```
fangting-alchemy, line 31: what happened to `source[0]`?

The comment should be updated too.

adamegyed: Previously, the layout was:
- length (4 bytes)
- index (1 bytes)
- content (variable length)

The new layout is:
- index (1 byte)
- length (4 bytes)
- content (variable)

Good call on updating comments, will go back and edit all of them

Zer0dot (approved):
Honestly looks great, neat way to increase efficiency and remove something we just didn't need!

PR #216, opened 2024-10-03
feat: convert ERC20TokenLimit module to prevalidation hook

fangting-alchemy:
## Motivation
`ERC20TokenLimitModule` currently provide execution hooks for validation functions.
This can be converted to validation hooks to revert before getting to execution phase. It also makes it possible to merge this module with other validation hook modules.

## Solution

Convert ERC20TokenLimit module to pre-validation hook from execution hooks.


adamegyed:
I think to get this PR to work, we'll also need to update how the ERC20TokenLimitModule is used in the gas benchmarks.

fangting-alchemy:
On second thought, this is not a good idea. Especially in the case where execution can be reverted while state changed in validation stays. Gonna close this PR.

PR #219, opened 2024-10-03
feat: hook mem management

adamegyed:
## Motivation

We can reduce storage loads and memory operations while running hooks.

## Solution

Get rid of the `PostExecToRun` struct, and replace the parameter to `_doCachedExecHook` with just `HookConfig[] memory hooks` and `bytes[] memory postHookDatas`.

Since we're not using the "flags" feature of LinkedListSetLib, we can omit that from the library (and its inner component steps, like clearing user flags and persisting the "has next" flag. Do this by copying in LinkedListSet, and porting over the unit and invariant test for it.

Define custom memory loaders for execution hooks, handling the 3 possible cases: selector-associated only, validation-associated only, and both types.

Update `_doPreHooks` to take in an externally allocated memory array, and to collect the post hook datas in the new format.

Introduce a new storage field within `validationData` to hold the lengths of each hook type. This helps avoid needless `sload`s during hook loading.
- Do not do the same for execution hooks, as it actually increases the number of `sload`s for native functions. With the general trend in ERC-6900 v0.8 of making better use of `execute` functions, there's less of a priority to optimize for module-defined execution functions.


Zer0dot:
Super cool, so I understand the change better, we're no longer allocating new memory for `postHooksToRun`, and instead passing the same hookConfig, yeah?

.forge-snapshots/ModularAccount_Runtime_UseSessionKey_Case1_Counter.snap
```diff
@@ -1 +1 @@
-89279
```
Zer0dot: Any idea why this increased so much?

.forge-snapshots/ModularAccount_Runtime_UseSessionKey_Case1_Token.snap
```diff
@@ -1 +1 @@
-122709
```
Zer0dot: Same here, I wonder what causes the increase

.forge-snapshots/ModularAccount_Runtime_UseSessionKey_Case1_Counter.snap
adamegyed: I'm pretty sure this is because as of this PR, validation hooks have two locations where length is stored:
- in validation data
- In the hook's length slot

While it was previously only the latter. So, it should be 1 more `sload`. But this is mostly just a hunch, we'd have to profile it more carefully to be sure.

The next PR in the stack removes the duplicate length storage, so the gas drops back down to roughly the same as before. (though surprisingly, still slightly above).

adamegyed:
> so I understand the change better, we're no longer allocating new memory for postHooksToRun, and instead passing the same hookConfig, yeah?

That's correct. It uses the hook flags to determine whether or not it needs to run.

src/libraries/MemManagementLib.sol
```diff
@@ -0,0 +1,170 @@
...
+            cursor = llsMap[cursor];
+
+            assembly ("memory-safe") {
+                mstore(add(hooks, mul(size, 0x20)), cursor)
+            }
+        }
+
+        assembly ("memory-safe") {
+            // Update the free memory pointer with the now-known length of the array.
+            mstore(0x40, add(hooks, mul(add(size, 1), 0x20)))
```
Zer0dot, line 72: Do you know if this mul gets optimized to a shl by 5? I think this is something the optimizer should do right?

src/libraries/MemManagementLib.sol
```diff
@@ -0,0 +1,170 @@
...
+        uint256 halfLen = len / 2;
+
+        for (uint256 i = 0; i < halfLen; ++i) {
+            uint256 j = len - i - 1;
+
+            (hooks[i], hooks[j]) = (hooks[j], hooks[i]);
+        }
+    }
+
+    function freezeFMP() internal pure returns (MemSnapshot snapshot) {
```
Zer0dot: Where are this and `restoreFMP()` used?

src/libraries/MemManagementLib.sol
```diff
@@ -0,0 +1,170 @@
...
+        // This is accomplished by first writing to memory after the free memory pointer,
+        // then updating the free memory pointer to cover the newly-allocated data.
+        // To the compiler, writes to memory after the free memory pointer are considered "memory safe".
+        // See https://docs.soliditylang.org/en/v0.8.22/assembly.html#memory-safety
+        // Stack variable lifting done when compiling with via-ir will only ever place variables into memory
+        // locations below the current free memory pointer, so it is safe to compile this library with via-ir.
+        // See https://docs.soliditylang.org/en/v0.8.22/yul.html#memoryguard
+        assembly ("memory-safe") {
+            // It is critical that no other memory allocations occur between:
+            // -  loading the value of the free memory pointer into `ret`
```
Zer0dot, line 32:
```suggestion
            // -  loading the value of the free memory pointer into `hooks`
```

Zer0dot: I know your comment says it's memory safe since we're writing after the fmp, but technically, can't we screw the optimizer up if it decides to lift something to memory after this and move the fmp outside the asm block?

src/libraries/MemManagementLib.sol
```diff
@@ -0,0 +1,170 @@
...
+            // -  updating the free memory pointer to point to the newly-allocated data, which is done after all
+            // the values have been written.
+            hooks := mload(0x40)
+        }
+
+        while (!LinkedListSetLib.isSentinel(cursor) && cursor != bytes32(0)) {
```
Zer0dot, line 38: In which scenario can the cursor be zero? When the lls is empty, so the sentinel points to nothing?

src/libraries/MemManagementLib.sol
```diff
@@ -0,0 +1,170 @@
...
+            selectors := selectorsSet
+        }
+
+        return selectors;
+    }
+
+    function reverseArr(HookConfig[] memory hooks) internal pure {
+        bytes32[] memory casted;
+
+        // Cast to bytes32[] to allow for in-place reversal
```
Zer0dot: What does this comment mean exactly? Which operation done in the bytes32 version of `reverseArr()` doesn't work for these types? Or is the goal not to repeat the same reversing code?

src/libraries/MemManagementLib.sol
```diff
@@ -0,0 +1,170 @@
...
+        // SetValue is internally a bytes30, and both bytes4 and bytes30 are left-aligned. This cast is safe so
+        // long as only bytes4 entries are added to the set.
+        assembly ("memory-safe") {
+            selectors := selectorsSet
+        }
+
+        return selectors;
+    }
+
+    function reverseArr(HookConfig[] memory hooks) internal pure {
```
Zer0dot, line 122:
```suggestion
    function reverseArr(HookConfig[] memory | bytes4[] memory values) internal pure {
```

You can just use a union typ- OH WAIT no you can't. One day though I'm confident we'll have this, this is me silently revolting.

src/account/AccountStorage.sol
```diff
@@ -35,6 +31,10 @@ struct ValidationData {
     bool isSignatureValidation;
     // Whether or not this validation is allowed to validate ERC-4337 user operations.
     bool isUserOpValidation;
+    // Length of the validation hooks array.
```
fangting-alchemy: Can we add why we want those two flags?

src/libraries/MemManagementLib.sol
adamegyed: No, stack variables that are lifted to memory go within the `memoryguard` allocated segment, which is between the zero word in memory (0x60) and user-allocated memory. Basically, if stack lifting occurs, the FMP will not start at `0x80`, but some greater value, and the space in between is where the lifted variables will be stored. 

Context: https://docs.soliditylang.org/en/latest/yul.html#memoryguard

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+bytes32 constant SENTINEL_VALUE = bytes32(uint256(1));
+
+struct LinkedListSet {
+    // Byte Layout
+    // | value | 0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA__ |
+    // | meta  | 0x______________________________________________________________BB |
+
+    // Bit-layout of the meta byte
+    // | unused   | 0b0000000_ |
+    // | sentinel | 0b_______A |
```
fangting-alchemy, line 19:
```suggestion
    // | sentinel | 0b_______1 |
```

howydev: since its a `bytes32` this should probably be `0b______01` right?

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+        mapping(bytes32 => bytes32) storage map = set.map;
+        bytes32 unwrappedKey = SetValue.unwrap(value);
+
+        bytes32 nextValue = map[unwrappedKey];
+        if (unwrappedKey == bytes32(0) || nextValue == bytes32(0)) return false;
+
+        bytes32 prevKey = SENTINEL_VALUE;
+        bytes32 currentVal;
+        do {
+            currentVal = map[prevKey];
```
howydev: really small nit, the 3 k/v pairs in sequence are:

prevKey -> prevVal/currKey -> currVal/nextKey -> nextVal

so by that, this var should be `prevVal` or `currKey`, right?

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+    // Bit-layout of the meta byte
+    // | unused   | 0b0000000_ |
+    // | sentinel | 0b_______A |
+
+    // Key excludes the meta bytes, except for the sentinel value, which is 0x1
+    mapping(bytes32 => bytes32) map;
+}
+
+/// @title Linked List Set Library
+/// @notice This library provides a set of functions for managing enumerable sets of bytes30 values.
```
howydev:
```suggestion
/// @notice This library provides a set of functions for managing enumerable sets of bytes31 values.
```

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+        } while (!isSentinel(currentVal) && currentVal != bytes32(0));
+        return false;
+    }
+
+    /// @notice Remove a value from a set, given the previous value in the set.
+    /// @dev This is an O(1) operation but requires additional knowledge.
+    /// @param set The set to remove the value from.
+    /// @param value The value to remove.
+    /// @param prev The previous value in the set.
+    /// @return True if the value was removed, false if the value does not exist.
```
howydev: (might need to run lint on length after)

```suggestion
    /// @return True if the value was removed, false if the value does not exist, or if the wrong prev was specified.
```

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+    /// @param set The set to remove the values from.
+    function clear(LinkedListSet storage set) internal {
+        mapping(bytes32 => bytes32) storage map = set.map;
+        bytes32 cursor = SENTINEL_VALUE;
+
+        do {
+            bytes32 next = map[cursor];
+            map[cursor] = bytes32(0);
+            cursor = next;
+        } while (!isSentinel(cursor) && cursor != bytes32(0));
```
howydev, line 129: sanity check for myself: cursor == `bytes32(0)` only in the case of running .clear on an empty set that never had a tryAdd operation right?

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+        // See https://docs.soliditylang.org/en/v0.8.22/yul.html#memoryguard
+        assembly ("memory-safe") {
+            // It is critical that no other memory allocations occur between:
+            // -  loading the value of the free memory pointer into `ret`
+            // -  updating the free memory pointer to point to the newly-allocated data, which is done after all
+            // the values have been written.
+            ret := mload(0x40)
+        }
+
+        while (!isSentinel(cursor) && cursor != bytes32(0)) {
```
howydev, line 179: Is there a concern here with these being put into memory as a solution to stack too deep?

There's only one concerning use of this which is in `toHookConfigArray` in `_doPreHooks`. If that parent function adds a bunch of stack vars, how does the compiler determine which vars to put into memory to solve stack too deep?

cc other yul chads @Zer0dot @fangting-alchemy for help

howydev: oh, addressed here: https://github.com/alchemyplatform/modular-account/pull/219/files/4b0fdca16bec11067c043219ac3c12139a2de375#r1788422100

src/account/AccountStorage.sol
adamegyed: The flags already have comments, we need to store state for `isGlobal`, `isSignatureValidation`, and `isUserOpValidation` because they're part of install params.

src/libraries/LinkedListSetLib.sol
```diff
@@ -0,0 +1,200 @@
...
+    // | unused   | 0b0000000_ |
+    // | sentinel | 0b_______A |
+
+    // Key excludes the meta bytes, except for the sentinel value, which is 0x1
+    mapping(bytes32 => bytes32) map;
+}
+
+/// @title Linked List Set Library
+/// @notice This library provides a set of functions for managing enumerable sets of bytes30 values.
+library LinkedListSetLib {
```
jaypaik, line 31: Can we use:

https://github.com/erc6900/modular-account-libs/blob/v0.8.0-rc.1/src/libraries/LinkedListSetLib.sol

src/libraries/LinkedListSetLib.sol
adamegyed: `0b` denotes binary, not hex, so it's only describing one byte. An underscore here means not used, and "A" here means it is the variable. If it is a sentinel, it will be 1, if it is not, then it will be 0. I think the way it currently is written is the most accurate.

src/libraries/LinkedListSetLib.sol
jaypaik: > Since we're not using the "flags" feature of LinkedListSetLib, we can omit that from the library (and its inner component steps, like clearing user flags and persisting the "has next" flag. Do this by copying in LinkedListSet, and porting over the unit and invariant test for it.

Just saw this. Hmm.

jaypaik: Can you make it clear in the NatSpec what was modified for this custom version of LinkedListSetLib?

.forge-snapshots/ModularAccount_Runtime_UseSessionKey_Case1_Counter.snap
Zer0dot: gotcha

src/libraries/MemManagementLib.sol
adamegyed: Yes, there are two possible empty states:
- Sentinel points to zero (uninitialized)
- Sentinel points to sentinel (initialized, but now empty).
from https://github.com/erc6900/modular-account-libs/blob/b63b1a872acdc92972a66f43dd17187f4ceafc97/src/libraries/LinkedListSetLib.sol#L206

src/libraries/MemManagementLib.sol
adamegyed: I was curious about this, so I stepped through it with the foundry debugger and yes, it does perform this optimization. I tested this with the `gas` profile, so with via-ir.

In general, I've found that the optimizer does fine with the low-level equivalence stuff when working with stack vars. It's memory that it is iffy with.

Screenshot 2024-10-08 at 12 06 37 PM

src/libraries/MemManagementLib.sol
adamegyed: They're not used... will remove for now.

src/libraries/MemManagementLib.sol
adamegyed: It also protects against a hypothetical risk where we step "off" of the actual linked list. That should only happen if there's a bug, or if the user does hacky stuff with an `upgradeTo` and/or `delegatecall` on a different implementation.

src/libraries/MemManagementLib.sol
adamegyed: Yeah the goal is just to not repeat code. I'm pretty sure this overload of the function flattens to a no-op in the IR optimization step.

I'll update the comment to be clearer that this is just to not repeat the code.

src/libraries/LinkedListSetLib.sol
adamegyed: Yes, this should be updated to currentKey. The distinction only really mattered when the LLS had flags, which caused there to be a difference between the key and the value (the key for the next element was the previous value with the flags removed).

Updated!

src/libraries/LinkedListSetLib.sol
adamegyed: Correct!

src/libraries/LinkedListSetLib.sol
adamegyed: Yep, will update the natspec. We can also revisit whether this omission actually results in a performance increase at the end of the PR stack, since the way it is used changes slightly.

src/account/AccountStorage.sol
adamegyed: Ah sorry, this is about the length fields, right? I'll expand the comment there.

src/libraries/MemManagementLib.sol
Zer0dot: Sick thanks for doing this, cool!

src/libraries/MemManagementLib.sol
Zer0dot: Didn't know about that, but just to rephrase what that means for us-- via-ir will allocate memory before the free memory pointer at 0x40, which it'll use for lifting stack variables if needed. Pretty cool!

src/libraries/MemManagementLib.sol
Zer0dot: Gotcha, thanks!

Zer0dot (approved):

PR #220, opened 2024-10-03
feat: move validation hooks to be LLS

adamegyed:
## Motivation

By using a storage array for pre-validation hooks, we can't hold the length of the pre-validation hooks in the same slot as the validation data. This leads to an extra `sload` happening during most execution paths, as two lengths are considered, even though they will always contain the same value.

## Solution

Convert validation hooks in storage to a LinkedListSet.


src/libraries/MemManagementLib.sol
```diff
@@ -109,6 +109,23 @@ library MemManagementLib {
         return hooks;
     }
 
+    // todo: can we save codesize by combining these two functions?
```
Zer0dot: I think you could pass the storage pointer and length as parameters to do this without a flag, or you could use a flag. Wdyt?

src/account/ModularAccountBase.sol
```diff
@@ -571,10 +571,16 @@ abstract contract ModularAccountBase is
 
         // Do preUserOpValidation hooks
         HookConfig[] memory preUserOpValidationHooks =
-            getAccountStorage().validationData[userOpValidationFunction].validationHooks;
+            MemManagementLib.loadValidationHooks(getAccountStorage().validationData[userOpValidationFunction]);
 
-        for (uint256 i = 0; i < preUserOpValidationHooks.length; ++i) {
-            (userOp.signature, signature) = signature.advanceSegmentIfAtIndex(uint8(i));
+        for (uint256 i = preUserOpValidationHooks.length; i > 0; i) {
+            // Decrement here, instead of in the loop body, to convert from length to an index.
```
Zer0dot, line 577: I guess this comment doesn't really work here since we're counting backwards and not indexing by "i" right, minor nit, fine with keeping it as is

src/account/ModularAccountBase.sol
```diff
@@ -611,12 +617,18 @@ abstract contract ModularAccountBase is
     ) internal {
         // run all preRuntimeValidation hooks
         HookConfig[] memory preRuntimeValidationHooks =
-            getAccountStorage().validationData[runtimeValidationFunction].validationHooks;
+            MemManagementLib.loadValidationHooks(getAccountStorage().validationData[runtimeValidationFunction]);
+
+        for (uint256 i = preRuntimeValidationHooks.length; i > 0;) {
+            // Decrement here, instead of in the loop body, to convert from length to an index.
```
Zer0dot: Same as before, but fine keeping as is

Zer0dot (approved):

src/libraries/MemManagementLib.sol
adamegyed: I tested merging them, and surprisingly it saves both gas and codesize. So let's go with that.

src/account/ModularAccountBase.sol
adamegyed: Well, we're counting by `i` for the `preUserOpValidationHooks` array, and by `preUserOpValidationHooks.length - i - 1` for the per-hook data. For clarity, I'll update the comment to be about decrementing as the first step of the loop body, rather than in the loop's updater step, to handle the length = 0 case.

src/account/ModularAccountBase.sol
adamegyed: also updated as above

PR #221, opened 2024-10-04
feat: add factory token sweeping

fangting-alchemy:
## Motivation
There is no way to sweep tokens in the factory contract.

## Solution
Add a method to sweep tokens in the factory contract.

### DIFF
--- a/test/factory/AccountFactory.t.sol
+++ b/test/factory/AccountFactory.t.sol
@@ -6,7 +6,11 @@ import {ModularAccount} from "../../src/account/ModularAccount.sol";
 import {AccountTestBase} from "../utils/AccountTestBase.sol";
 import {TEST_DEFAULT_VALIDATION_ENTITY_ID} from "../utils/TestConstants.sol";
 
+import {MockERC20} from "../mocks/MockERC20.sol";
+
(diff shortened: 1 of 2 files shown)


test/factory/AccountFactory.t.sol
```diff
@@ -35,4 +39,14 @@ contract AccountFactoryTest is AccountTestBase {
         // Assert the return addresses are the same
         assertEq(address(account), address(account2));
     }
+
+    function test_withdraw() public {
```
adamegyed, line 43: nit: we should also test the ability to withdraw native tokens

adamegyed (approved):
Looks good, would like 1 small extra test case for native token transfers

test/factory/AccountFactory.t.sol
fangting-alchemy: It is actually impossible to have native tokens in contract.

During sync, I mentioned one possibility: addStake, then withdrawStake to factory.
Turns out it will fail to withdraw since the factory will reject receiving ETH without a payable fallback or receive func. 

Gonna still keep the code in in case there is a missed hack.

PR #222, opened 2024-10-04
feat: user op call buffers

adamegyed:
## Motivation

When the account performs external calls for user op validation (IValidationHookModule.preUserOpValidationHook, IValidationmodule.validateUserOp), the account must re-encode the calls each time, which repeats a lot of memory allocations. Logically, these calls are very similar, only differing in target address, entity id, and signature segment. We can save runtime gas by only allocating memory once, and re-using it for each call, updating the fields as needed.

## Solution

Implement that optimization by adding  functions to ExecutionLib to:
- Allocate the call buffer
- Update the selector
- Perform each call.

Add fuzz testing for a varying number of pre-validation hooks, and with varying length and content for each


.forge-snapshots/ModularAccount_UserOp_UseSessionKey_Case1_Token.snap
```diff
@@ -1 +1 @@
-239176
\ No newline at end of file
+232816
```
howydev: theres 7k in savings just from memory allocation of 2 additional calls? thats crazy

src/helpers/EmptyCalldataSlice.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: UNLICENSED
```
howydev, line 1: nit: this seems to be used only in `SparseCalldataSegmentLib`, could we move it there?

adamegyed: In the later PRs in the stack, we have to use it elsewhere too, in ModularAccountBase.

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+
+            let actualCallLength := add(userOpSigRelativeOffset, 0x84)
+
+            // Add in the new signature length, with word alignment. This is safe to do because the signature
+            // segment length is guaranteed to be less than the size of the previous entire signature length.
+            actualCallLength := add(actualCallLength, and(add(signatureSegment.length, 0x1f), not(0x1f)))
+
+            // Perform the call, reverting on failure or insufficient return data.
+            switch and(
+                gt(returndatasize(), 0x1f),
```
howydev, line 187: i assume `call` always happens first before `returndatasize` here?

adamegyed: Correct, yul evaluates these expression right-to-left. There's a similar pattern done in solmate and solady.

adamegyed: We should add a test for this case though, to be sure.

.forge-snapshots/ModularAccount_UserOp_UseSessionKey_Case1_Token.snap
adamegyed: I think part of the reason it adds so much is because of how the `PackedUserOp` struct has a large number of sub-components (9), which all affect the encoding logic.

src/account/ModularAccountBase.sol
```diff
@@ -562,42 +562,50 @@ abstract contract ModularAccountBase is
...
+        bytes calldata signature
     ) internal returns (uint256) {
         uint256 validationRes;
 
         // Do preUserOpValidation hooks
         HookConfig[] memory preUserOpValidationHooks =
             MemManagementLib.loadValidationHooks(getAccountStorage().validationData[userOpValidationFunction]);
 
+        UOCallBuffer userOpCallBuffer;
+        if (!_validationIsNative(userOpValidationFunction) || preUserOpValidationHooks.length > 0) {
```
Zer0dot, line 574: In an ideal world this would optimize out entirely in the standard ModularAccount. I wonder if it does, either way I don't see a better way to go about it.

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+            let moduleAddress := shr(96, moduleEntity)
+
+            // Update the buffer with the entity Id
+            mstore(add(buffer, 0x24), entityId)
+
+            // Get the offset of the user op signature in the buffer.
+            // The PackedUserOperation starts at the 5th word in the buffer (0x20 * 4 = 0x80).
+            // It is the 9th element in PackedUserOp (so add 0x20 * 8 = 0x100 to the buffer start).
+            // So we start at 0x184, to include the selector length.
+            // THen, to convert from a relative to an absolute offset, we need to add the buffer start, selector,
```
```suggestion
            // Then, to convert from a relative to an absolute offset, we need to add the buffer start, selector,
```

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+            // and to skip over the entityId, offset, and hash.
+            let userOpSigRelativeOffset := mload(add(buffer, 0x184))
+            let userOpSigAbsOffset := add(add(buffer, userOpSigRelativeOffset), 0x84)
+
+            // Copy in the signature segment
+            // Since the buffer's copy of the signature exceeds the length of any sub-segments, we can safely write
```
Zer0dot, line 161: In other words, the offset is after all other values in the buffer, yeah?

adamegyed: Are you asking about how this computes `userOpSigAbsOffset`? If so, it's about converting from the offset value in the buffer itself, which describes where the field is relative to the other fields in the ABI-encoded bytes, to a memory pointer. 

The `userOp.signature` field is twice nested with offsets:
- PackedUserOperation is a struct, which is encoded as an offset + tuple
- signature is a `bytes` type, which is encoded as an offset + length + contents

For the first offset of the userOp itself, we don't actually need to follow the offset lengths, because it will always be the next thing encoded after the outermost params (entityId, userOp, userOpHash). This is because the previous `abi.encodeCall` will always encode in strict mode, and that is why we add a fixed value of `0x84`.

We don't know a constant value for how far into the buffer the signature starts, because it will be encoded after some other dynamic-length types (`userOp.initCode`, `userOp.callData`, `userOp.paymasterAndData`). So, we need to actually read and follow the offsets.

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+            // over it.
+            mstore(userOpSigAbsOffset, signatureSegment.length)
+            // If there is a nonzero signature segment length, copy in the data.
+            if signatureSegment.length {
+                // Because we will be sending the data with word-aligned padding ("strict ABI encoding"), we need
```
Zer0dot, line 166: Is this because the signature has to be a multiple of 32 bytes, so we risk passing junk data if it's shorter and we don't zero it out?

If so, what's the risk with just passing the junk data?

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+
+            // Add in the new signature length, with word alignment. This is safe to do because the signature
+            // segment length is guaranteed to be less than the size of the previous entire signature length.
+            actualCallLength := add(actualCallLength, and(add(signatureSegment.length, 0x1f), not(0x1f)))
+
+            // Perform the call, reverting on failure or insufficient return data.
+            switch and(
+                gt(returndatasize(), 0x1f),
+                call(
+                    // If gas is the leftmost item before the call, it *should* be placed immediately before the
```
Zer0dot, line 189: This will inevitably be tested in integration testing regardless since this flow is super common, but adding this comment as a mental note if we end up having a validation error and going through a bunch of pain to debug it-- we should look here.

src/libraries/ExecutionLib.sol
adamegyed: If the module this is sent to follows the solidity principle of "strict encoding out, loose decoding in", then it is ok to send garbage additional data. I included this here to keep the external behavior of the account exactly the same as before with these optimizations, to make it easier to demonstrate "correctness". E.g. in the tests, I assert that the data sent is equivalent to an `abi.encodeCall(...)` with the expected data.

While we technically can omit this, I'd prefer not to, to make the behavior of the account towards modules more in line with what a standard 6900 account would do.

src/libraries/ExecutionLib.sol
Zer0dot: Thanks for the chat, and this explanation makes sense! Leaving this unresolved as it may help other reviewers too!

src/libraries/ExecutionLib.sol
Zer0dot: Agreed, makes sense to me!

src/helpers/EmptyCalldataSlice.sol
```diff
@@ -0,0 +1,13 @@
...
+pragma solidity ^0.8.26;
+
+function getEmptyCalldataSlice() pure returns (bytes calldata) {
+    bytes calldata empty;
+
+    assembly ("memory-safe") {
```
Zer0dot, line 7: What would happen if we omit this?

Tried on remix, turns out it's an error as it would lead to undefined behaviour. I assume that if it wasn't an error, this would just be a zero offset pointing to the first word in calldata. So it's "length" would be the first word. Interesting.

test/account/UOCallBuffer.t.sol
```diff
@@ -0,0 +1,238 @@
...
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+
+import {HookConfigLib} from "../../src/libraries/HookConfigLib.sol";
+import {ModuleEntity, ModuleEntityLib} from "../../src/libraries/ModuleEntityLib.sol";
+import {ValidationConfigLib} from "../../src/libraries/ValidationConfigLib.sol";
+
+import {MockModule} from "../mocks/modules/MockModule.sol";
+import {AccountTestBase} from "../utils/AccountTestBase.sol";
+
+contract UOCallBufferTest is AccountTestBase {
```
Zer0dot, line 16: Do you want to add withSMATest() modifiers to these?

test/account/UOCallBuffer.t.sol
```diff
@@ -0,0 +1,238 @@
...
+        vm.prank(address(entryPoint));
+        account1.validateUserOp(userOp, userOpHash, 1 wei);
+    }
+
+    function testFuzz_variableLengthUOCalls(
+        uint8 preValidationHookCount,
+        bytes[254] memory preValidationHookData,
+        bytes memory validationData
+    ) public {
+        preValidationHookCount = uint8(bound(preValidationHookCount, 0, 254));
```
Zer0dot, line 124: Is this equivalent to vm.assume?

adamegyed: Yes, but `bound` is preferred over `assume` for fuzzing because they behave slightly differently. `assume` will simply discard a fuzz run if the expression is false, and `bound` will instead modify the incoming value to be within the bounded range (just using a modulo). More info here: https://book.getfoundry.sh/reference/forge-std/bound

src/helpers/EmptyCalldataSlice.sol
adamegyed: Yep, the compiler doesn't like uninitialized `calldata` reference types.

I did find something new with testing: we can omit setting the offset, and just set the length, to save 2 gas.

Will update and push the change.

src/account/ModularAccountBase.sol
adamegyed: Confirming that the memory check for `preUserOpValidationHooks.length > 0` gets optimized out in ModularAccount.

src/libraries/ExecutionLib.sol
adamegyed: Added a test for short return data!

test/account/UOCallBuffer.t.sol
adamegyed: Yep, added!

test/account/UOCallBuffer.t.sol
Zer0dot: Sick, I wasn't aware assume just wrecks your fuzz run, I guess it'd also force a re-run, so it sounds super painful.

Zer0dot (approved):

test/account/UOCallBuffer.t.sol
adamegyed: It's not too bad in general, and it's preferred for one-off exclusions (things like `!= 0`), but if too many of the generated test cases fall into the `assume` skip branch, then the fuzzer will report a failure.

.forge-snapshots/ModularAccount_Runtime_UseSessionKey_Case1_Token.snap
```diff
@@ -1 +1 @@
-122994
\ No newline at end of file
+123321
```
fangting-alchemy: Why costing more here (same for SMA too not surprising)?

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+
+    // Allocate a buffer to call user op validation and validation hook functions. Both of these take the form of
+    // - bytes4 selector
+    // - uint32 entityId
+    // - PackedUserOperation userOp
+    // - bytes32 userOpHash
+    // The buffer starts with the selector for `preUserOpValidationHook`, and can be updated later to
+    // `validateUserOp`. When perfomring the actual function calls later, update the entityId field and selector,
+    // as as needed.
+    function allocateUserOpCallBuffer(PackedUserOperation calldata userOp, bytes32 userOpHash)
```
fangting-alchemy: nit:
```suggestion
    function allocateUOValidationHookCallBuffer(PackedUserOperation calldata userOp, bytes32 userOpHash)
```

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+    function convertToValidationBuffer(UOCallBuffer buffer) internal pure {
+        // Selector is treated as a uint32 to be right-aligned in the word.
+        uint32 selector = uint32(IValidationModule.validateUserOp.selector);
+
+        assembly ("memory-safe") {
+            // We want to write in the selector without writing over anything else in the buffer, so we save the
+            // length, write over a portion of the length, and restore it.
+            let bufferLength := mload(buffer)
+            mstore(add(buffer, 4), selector)
+            mstore(buffer, bufferLength)
```
fangting-alchemy, line 133: Why not just overwrite the selector without writing over length and restore?

fangting-alchemy: nvm, it will write over 32 bytes.

src/libraries/ExecutionLib.sol
```diff
@@ -69,4 +82,137 @@ library ExecutionLib {
...
+    // Invokes either a user op validation hook, or validation function.
+    function invokeUserOpCallBuffer(
+        UOCallBuffer buffer,
+        ModuleEntity moduleEntity,
+        bytes calldata signatureSegment
+    ) internal returns (uint256 validationData) {
+        assembly ("memory-safe") {
+            // Load the module address and entity Id
+            let entityId := and(shr(64, moduleEntity), 0xffffffff)
+            let moduleAddress := shr(96, moduleEntity)
```
fangting-alchemy, line 146: nit: consistency using decimals and hexadecimal.

fangting-alchemy (approved):

.forge-snapshots/ModularAccount_Runtime_UseSessionKey_Case1_Token.snap
adamegyed: I think this is coming from extra calldata->memory copying, because this PR changes the return type of `SparseCalldataSegmentLib.advanceSegmentIfAtIndex` from `bytes memory` to `bytes calldata`, and it can't short-circuit with the 0-length memory case anymore (I think, not 100% sure).

But, the changes in the next PR reduce this metric's gas cost to before even the starting one in this PR!

src/libraries/ExecutionLib.sol
adamegyed: I've been trying to keep a convention of using decimal when describing bits within a single word, and hex when describing bytes in memory.

src/libraries/ExecutionLib.sol
adamegyed: Will update to `allocateUOValidationCallBuffer`, excluding "Hook", because it's the same buffer used for both validation hooks and the validation itself.

PR #223, opened 2024-10-04
feat: add SMA known selectors

Zer0dot:
## Motivation

We currently don't have a way to check if a selector is native for SMA selectors.

## Solution

This PR adds a new `SemiModularKnownSelectorsLib` library which implements only the `isNativeFunction()` function, which then checks if the selector is a native selector in the `KnownSelectorsLib` library or a native SMA selector. Note that the initializer is not included-- just like in the standard `KnownSelectorsLib`.

### DIFF
--- a/src/libraries/SemiModularKnownSelectorsLib.sol
+++ b/src/libraries/SemiModularKnownSelectorsLib.sol
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+import {SemiModularAccountBase} from "../account/SemiModularAccountBase.sol";
+import {KnownSelectorsLib} from "./KnownSelectorsLib.sol";
+
+/// @dev Library to help to check if a selector is a know function selector of the SemiModularAccountBase or
+/// ModularAccount contract
+library SemiModularKnownSelectorsLib {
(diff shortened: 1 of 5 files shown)


src/libraries/SemiModularKnownSelectorsLib.sol
```diff
@@ -0,0 +1,18 @@
...
+pragma solidity ^0.8.26;
+
+import {SemiModularAccountBase} from "../account/SemiModularAccountBase.sol";
+import {KnownSelectorsLib} from "./KnownSelectorsLib.sol";
+
+/// @dev Library to help to check if a selector is a know function selector of the SemiModularAccountBase or
+/// ModularAccount contract
+library SemiModularKnownSelectorsLib {
+    function isNativeFunction(bytes4 selector) internal pure returns (bool) {
+        return KnownSelectorsLib.isNativeFunction(selector)
```
fangting-alchemy, line 11: Nit: maybe rename `isNativeFunction` to `isModularAccountNativeFunction`?

fangting-alchemy (approved):

src/libraries/SemiModularKnownSelectorsLib.sol
Zer0dot: Technically it also checks IAccount, UUPS, etc-- I think it's probably fine as is now, but happy to discuss this further later on!

PR #224, opened 2024-10-04
feat: change sig validation to switch off 1 byte of calldata

howydev:
We currently use OZ.SignatureChecker to switch between doing ECDSA and contract owner signature verification. In newer versions, it switches based on an `extcodesize` on the owner's address.

Given that 7702 would cause addresses to be able to behave like EOAs and accounts, it's safer to use a scheme that specifies which signature validation mode is used. This PR changes the scheme in our modules and in SMA to use the one in LightAccount v2


src/account/SemiModularAccountBase.sol
```diff
@@ -145,14 +143,39 @@ abstract contract SemiModularAccountBase is ModularAccountBase {
...
             return _1271_INVALID;
         }
         return super._exec1271Validation(sigValidation, hash, signature);
     }
 
+    function _checkSignature(address owner, bytes32 digest, bytes memory sig) internal view returns (bool) {
+        if (sig.length < 1) {
+            revert InvalidSignatureType();
+        }
+        uint8 sigType = uint8(bytes1(sig));
```
adamegyed: Could we make this an enum, like we do for LightAccount? https://github.com/alchemyplatform/light-account/blob/develop/src/common/BaseLightAccount.sol#L17

src/account/SemiModularAccountBase.sol
```diff
@@ -145,14 +143,39 @@ abstract contract SemiModularAccountBase is ModularAccountBase {
...
+        // remove first byte from sig in memory
+        assembly ("memory-safe") {
+            let sigLen := mload(sig)
+            sig := add(sig, 1)
+            mstore(sig, sub(sigLen, 1))
```
adamegyed, line 165: This technically modifies the `PackedUserOperation memory` variable, which would make the memory struct not safe to use after the modification. Thankfully, we don't use it after this step, because all pre-validation hooks have run before.

This will also slightly change with UO call buffers, but that should make it easier, because this function will get the signature in a `bytes calldata` type.

src/modules/validation/ECDSAValidationModule.sol
```diff
@@ -3,6 +3,8 @@ pragma solidity ^0.8.26;
 
 import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
 import {IERC165} from "@openzeppelin/contracts/interfaces/IERC165.sol";
+
```
adamegyed: nit: no extra whitespace here

test/utils/ModuleSignatureUtils.sol
```diff
@@ -25,6 +25,8 @@ contract ModuleSignatureUtils {
 
     uint8 public constant SELECTOR_ASSOCIATED_VALIDATION = 0;
     uint8 public constant GLOBAL_VALIDATION = 1;
+    uint8 public constant EOA_TYPE_SIGNATURE = 0;
+    uint8 public constant CONTRACT_OWNER_TYPE_SIGNATURE = 1;
```
adamegyed: Do we use this variable anywhere? If we don't, we should add tests for contract owner sigs eventually (doesn't have to be in this PR).

adamegyed (commented):
Looks good, small comments

src/account/SemiModularAccountBase.sol
howydev: good catch!

howydev: yeah would be nice to use a calldata slice for this

test/utils/ModuleSignatureUtils.sol
howydev: currently not, no, will add once all the contract sig stuff are done

howydev: removed this var till then to fix

src/account/SemiModularAccountBase.sol
howydev: fixed

src/libraries/SignatureType.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity ^0.8.26;
+
+enum SignatureType {
```
adamegyed: Not blocking, but maybe this doesn't belong in `libraries/`? I can't really think of a great place for it haha. Maybe in `helpers/`? Will leave it up to you to decide where to put it.

adamegyed (approved):
Looks good, 1 callout to consider but leaving it up to you

src/libraries/SignatureType.sol
jaypaik: I like `helpers/`.

huaweigu:
Is OZ aware of this issue? This change seems technically sound, but could it introduce compatibility issues between different vendors (wallets, SDKs, etc.)? Should we look into standardizing signature encoding?

PR #225, opened 2024-10-05
feat: runtime call buffers

adamegyed:
## Motivation

When the account performs external calls for runtime validation (IValidationHookModule.preRuntimeValidationHook, IValidationmodule.validateRuntime), the account must re-encode the calls each time, which repeats a lot of memory allocations. Logically, these calls are very similar, only differing in target address, entity id, account, and authorization segment. We can save gas by only allocating memory once, and re-using it for each call, updating the fields as needed.

## Solution

Implement that optimization by adding functions to ExecutionLib to:

Allocate the call buffer
Update the selector and/or the `account` field.
Perform each call.
Add fuzz testing for a varying number of pre-validation hooks, and with varying length and content for each


src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            IValidationModule.validateRuntime,
+            (address(0), uint32(0), msg.sender, msg.value, callData, authorization)
+        );
+
+        assembly ("memory-safe") {
+            result := buffer
+        }
+
+        // Prepare the buffer for pre-runtime validation hooks.
+        ExecutionLib.prepareRuntimeCallBufferPreValidationHooks(result);
```
Zer0dot: Technically we can omit the `ExecutionLib.` here right?

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+    function prepareRuntimeCallBufferPreValidationHooks(RTCallBuffer buffer) internal pure {
+        uint32 selector = uint32(IValidationHookModule.preRuntimeValidationHook.selector);
+        assembly ("memory-safe") {
+            // Update the buffer with the selector. This will squash a portion of the `account` param for runtime
+            // validation, but that will be restored before calling.
+            mstore(add(buffer, 0x24), selector)
+
+            // Fix the calldata offsets of `callData` and `authorization`, due to excluding the `account` field.
+
+            // The offset of calldata starts out as 0x0c0, but for pre-validation hooks, it should be 0x0a0.
```
Zer0dot: This is because we're skipping the first word (account) so we need to reduce the offset by 0x20, yeah?

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            mstore(add(buffer, 0xa4), 0xa0)
+
+            // The offset of authorization should be decremented by one word.
+            let authorizationOffsetPtr := add(buffer, 0xc4)
+            // Get the stored value. This wil lbe wrong for preRuntimeValidationHooks, because the buffer size is
```
```suggestion
            // Get the stored value. This will be wrong for preRuntimeValidationHooks, because the buffer size is
```

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // buffer length - word-align(oldAuthorization length) + word-align(newAuthorization length) - 0x20 (to
+            // skip `account`),
+            // This is equivalent to:
+            // 4 (selector length) + 0x20 (authorization length field) + authorizationRelativeOffset +
+            // word-align(newAuthorization length)
+
+            let actualCallLength := add(authorizationRelativeOffset, 0x24)
+
+            // Add in the new authorization length, with word alignment. This is safe to do because the
+            // authorization
```
Zer0dot: Looks like the formatter chose violence today, might be worth just bringing the line below this up and re-formatting to get a nicer looking comment. Mega small nit though

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // Fix the calldata offsets of `callData` and `authorization`, due to including the `account` field for
+            // runtime validation.
+
+            // The offset of calldata should be reset back to 0x0c0. For pre-validation hooks, it was set to 0x0a0.
+            mstore(add(buffer, 0xa4), 0xc0)
+
+            // Get the offset of the authorization in the buffer.
+            // The authorization offset is the 6th word in the buffer (0x20 * 5 = 0xa0).
+            let authorizationOffsetPtr := add(buffer, 0xc4)
+            // Get the stored value. This will be the editted value for preRuntimeValidationHooks.
```
```suggestion
            // Get the stored value. This will be the edited value for preRuntimeValidationHooks.
```

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            let authorizationRelativeOffset := mload(authorizationOffsetPtr)
+            // Correct the authorization relative offset
+            authorizationRelativeOffset := add(authorizationRelativeOffset, 0x20)
+            // Fix the stored offset value by adding 0x20.
```
Zer0dot, line 365: I think this comment and the one above may be in reverse order

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // The offset of calldata should be reset back to 0x0c0. For pre-validation hooks, it was set to 0x0a0.
+            mstore(add(buffer, 0xa4), 0xc0)
+
+            // Get the offset of the authorization in the buffer.
+            // The authorization offset is the 6th word in the buffer (0x20 * 5 = 0xa0).
+            let authorizationOffsetPtr := add(buffer, 0xc4)
+            // Get the stored value. This will be the editted value for preRuntimeValidationHooks.
+            let authorizationRelativeOffset := mload(authorizationOffsetPtr)
+            // Correct the authorization relative offset
+            authorizationRelativeOffset := add(authorizationRelativeOffset, 0x20)
```
Zer0dot: This is because in `allocateRuntimeCalBuffer()` we call `prepareRuntimeCallBufferPreValidationHooks()` where the offset is decremented by 0x20, right?

So technically we're being a bit wasteful if a validation has no validation hooks, yeah?

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // Fix the stored offset value by adding 0x20.
+            mstore(authorizationOffsetPtr, authorizationRelativeOffset)
+
+            // Convert to an absolute offset
+            // Add the lengths of the selector, buffer length field, and the account field.
```
Zer0dot: I think the account field is already included in the relative offset, so we might need to edit this comment.
```suggestion
            // Add the lengths of the selector and buffer length field field.
```

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // Since the buffer's copy of the authorization exceeds the length of any sub-segments, we can safely
+            // write over it.
+            mstore(authorizationAbsOffset, authorizationSegment.length)
+            // If there is a nonzero authorization segment length, copy in the data.
+            if authorizationSegment.length {
+                // Because we will be sending the data with word-aligned padding ("strict ABI encoding"), we need
+                // to zero out the last word of the buffer to prevent sending garbage data.
+                let roundedDownAuthorizationLength := and(authorizationSegment.length, not(0x1f))
+                mstore(add(authorizationAbsOffset, add(roundedDownAuthorizationLength, 0x20)), 0)
+                calldatacopy(
```
Zer0dot, line 385:
```suggestion
                // Copy the authorization segment from calldata into the correct location in the buffer.
                calldatacopy(
```

Wdyt of adding a comment here?

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
```
Zer0dot, line 290: Didn't comment here, but it might be worth adding a smol comment for this calldatacopy:
```suggestion
                // Copy the authorization segment from calldata into the correct location in the buffer.
                calldatacopy(
```

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // skip `account`),
+            // This is equivalent to:
+            // 4 (selector length) + 0x20 (authorization length field) + authorizationRelativeOffset +
+            // word-align(newAuthorization length)
+
+            let actualCallLength := add(authorizationRelativeOffset, 0x24)
+
+            // Add in the new authorization length, with word alignment.
+            // This is safe to do because the authorization segment length is guaranteed to be less than the size
+            // of
```
Zer0dot: Comment formatting is weird here, but I think it's probably best to not waste time with the comment formatter for now and just clean it up later.

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // the previous entire authorization length.
+            actualCallLength := add(actualCallLength, and(add(authorizationSegment.length, 0x1f), not(0x1f)))
+
+            // Before performing the call, we need to check that the module has code.
+            // IValidationModule.validateRuntime has no return value, so an EOA added as a validation (perhaps for
+            // direct call validation) would authorize all calls, which is unsafe. Solidity inserts this check by
+            // default, but when we're making calls manually via call buffers, we need to do the check ourselves.
+            if iszero(extcodesize(moduleAddress)) {
+                // todo: do we want to create a custom error for this case?
```
Zer0dot: We couuuuuuld, but it's such a weird edge case. I don't think it's worth the extra PUSH4 PUSH0 MSTORE PUSH1 (that's what we'd have to add, yeah?) this is a "might bite us in the ass in the future if we don't do it" type statement.

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // 4 (selector length) + 0x20 (authorization length field) + authorizationRelativeOffset +
+            // word-align(newAuthorization length)
+
+            let actualCallLength := add(authorizationRelativeOffset, 0x24)
+
+            // Add in the new authorization length, with word alignment.
+            // This is safe to do because the authorization segment length is guaranteed to be less than the size
+            // of
+            // the previous entire authorization length.
+            actualCallLength := add(actualCallLength, and(add(authorizationSegment.length, 0x1f), not(0x1f)))
```
Zer0dot, line 402: Small question, is the length at the buffer offset `mload(buffer)` ever actually used? If not, we could consider using it, wdyt?

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+                    0
+                )
+        }
+
+        if (!success) {
+            revert RuntimeValidationFunctionReverted(moduleAddress, entityId, collectReturnData());
+        }
+    }
+
+    // Todo: this will need to change to accepting a PreHookBuffer in the future.
```
Zer0dot: ⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢠⣆⢀⣶⡶⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⣿⢸⠟⣠⣶⡷⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⢀⣀⠀⢀⣠⠴⠴⠶⠚⠿⠿⠾⠭⣤⣀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⢀⠴⢋⡽⠚⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠈⠙⠢⣀⠀⠀⠀⠀⠀⠀
⠀⠀⢀⡔⠁⡰⠋⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣴⠚⠛⣖⠀⠀⠀⠀
⠀⢀⡏⠀⡼⢡⠚⡛⠒⣄⠀⠀⠀⠀⣠⠖⠛⠛⠲⡄⠐⢯⠁⠀⠀⠹⡧⠀⠀⠀
⠀⣸⠀⠀⡇⠘⠦⣭⡤⢟⡤⠤⣀⠀⠣⣀⡉⢁⣀⠟⠀⠀⢷⠀⠀⠀⠙⣗⠀⠀
⠁⢻⠀⠀⢷⢀⡔⠉⢻⡅⣀⣤⡈⠙⠒⠺⠯⡍⠁⠀⠀⠀⢸⡆⠀⠀⠀⠘⡶⠄
⠀⣈⣧⠴⠚⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣰⠃⠀⠀⠀⠀⣸⡇⠀⠀⠀⠀⠸⣔
⣾⠋⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣿⣧⣤⡤⠴⠖⠋⢹⠃⠀⠀⠀⠀⠀⣷
⣿⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⣻⠁⠀⠀⠀⠀⠀⣿⠀⠀⠀⠀⠀⠀⣼
⠙⠑⣤⣀⠀⠀⠀⠀⠀⢀⠀⠀⢄⣐⠴⠋⠀⠀⠀⠀⠀⠀⠘⢆⠀⠀⠀⠀⣰⠟
⠀⠀⠀⣑⡟⠛⠛⠛⠛⠛⠛⠉⠉⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢴⡾⠋⠀
⠀⠀⠀⡾⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠘⡇⠀⠀
⠀⠀⣰⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡇⠀⠀
⠀⠀⠸⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢰⠃⠀⠃

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+            // would cause the relative offset of calldata to change. So, when loading the callData for self-exec,
+            // we must not load the relative calldata offset, and must instead use an absolute offset from the
+            // start of the buffer. Using an absolute offset is safe because the abi encoder will only generate
+            // “strict encoding mode” encodings, so it is guaranteed to be in that location.
+
+            bytes memory callData;
+
+            assembly ("memory-safe") {
+                // Get the memory address of the length of callData in the buffer.
+                // Because we don't know whether this will be invoked with the RTCallBuffer still as a
```
Zer0dot, line 452: Small unfinished comment, I'm not sure what it's supposed to say-- it seems like regardless of whether we're going to call a hook or a validation this offset would be the same from the start of the buffer, no?

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,267 @@ library ExecutionLib {
...
+                callData := add(buffer, 0xe4)
+            }
+
+            return callData;
+        } else {
+            // No buffer exists yet, just copy the data to memory and return it.
+            return data;
```
Zer0dot, line 459: Is there any case other than a direct call validation/fallback validation where there's no hook where this would be the code path?

Just curious.

test/account/RTCallBuffer.t.sol
```diff
@@ -0,0 +1,219 @@
...
+import {IValidationModule} from "@erc6900/reference-implementation/interfaces/IValidationModule.sol";
+
+import {HookConfigLib} from "../../src/libraries/HookConfigLib.sol";
+import {ModuleEntity, ModuleEntityLib} from "../../src/libraries/ModuleEntityLib.sol";
+import {ValidationConfigLib} from "../../src/libraries/ValidationConfigLib.sol";
+
+import {MockModule} from "../mocks/modules/MockModule.sol";
+import {AccountTestBase} from "../utils/AccountTestBase.sol";
+
+contract RTCallBufferTest is AccountTestBase {
```
Zer0dot, line 15: I think we're missing the `withSMATest` modifiers too here

src/libraries/ExecutionLib.sol
adamegyed: yep, the relative offsets have to change.

src/libraries/ExecutionLib.sol
adamegyed: Yep, I'll convert this to a private function instead and omit the library name.

src/libraries/ExecutionLib.sol
adamegyed: fixed

src/libraries/ExecutionLib.sol
adamegyed: Fixed!

src/libraries/ExecutionLib.sol
adamegyed: Yes, it is a bit wasteful, but 1) I don't really want to re-implement the `abi.encode...` logic by hand, and 2) tracking the various states of whether a buffer is allocated and whether it is currently in pre-validation hook mode or runtime validation mode is a bit tricky.

src/libraries/ExecutionLib.sol
adamegyed: Yep, fixed!

src/libraries/ExecutionLib.sol
adamegyed: I thought about this a bit more, but reverting with 0 data is actually the solidity builtin behavior for this case. Deets: https://alchemyinsights.slack.com/archives/C02VDQV5TKQ/p1710514180929839

So I think we're good to leave it as-is. I'll remove my comment.

src/libraries/ExecutionLib.sol
adamegyed: I don't think it is used. I previously tried the other approach of calculating the call length as`buffer length - old authorization length (rounded) + new authorization length (rounded)`, but unfortunately that was more gas expensive than the current solution, which uses the authorization's relative offset as the starting point for the length calculation.

src/libraries/ExecutionLib.sol
adamegyed: Ironically, I ended up not doing that, because tracking which type to use was more expensive than just always using the RTCallBuffer. Removing this comment.

src/libraries/ExecutionLib.sol
adamegyed: My comment was trying to say the realization that you said: "regardless of whether we're going to call a hook or a validation this offset would be the same from the start of the buffer".

While working on this, at first I mistakenly tried to follow the offsets to get to the start of callData, but there could be cases where there were pre-val hooks but no validation run (from SMA), as well as no pre-val hooks but an RT validation buffer call. So instead of tracking the usage state, I just switched it out to the fixed offset, because it's simpler + always correct.

Any suggestions for how to fix the comment?

src/libraries/ExecutionLib.sol
adamegyed: I think it is just those two cases: Direct call validation with no hooks, or SMA with no hooks.

test/account/RTCallBuffer.t.sol
adamegyed: Added!

src/libraries/ExecutionLib.sol
Zer0dot: Sounds like there's no way to make use of it without re-implementing the encodeCall manually, and we'd only save 1 word of memory, yeah?

adamegyed: Yes, it seems like it.

src/libraries/ExecutionLib.sol
Zer0dot:
```suggestion
                // This is always at a fixed offset from the start of the buffer.
```
Wdyt?

src/libraries/ExecutionLib.sol
Zer0dot: Good stuff!

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,264 @@ library ExecutionLib {
             }
         }
     }
+
+    function allocateRuntimeCallBuffer(bytes calldata callData, bytes calldata authorization)
```
fangting-alchemy: nit: 
```suggestion
    function allocateRuntimeValidationCallBuffer(bytes calldata callData, bytes calldata authorization)
```

src/libraries/ExecutionLib.sol
```diff
@@ -215,4 +222,264 @@ library ExecutionLib {
...
+                )
+        }
+
+        if (!success) {
+            revert PreRuntimeValidationHookFailed(moduleAddress, entityId, collectReturnData());
+        }
+    }
+
+    // Note: we need to add an extra check for codesize > 0 on the module, otherwise EOAs added as runtime
+    // validation would authorize all calls.
```
fangting-alchemy, line 333: How would 7702 delegated EOAs affect this? They would be able to bypass this check.

adamegyed: Two things:
1. With 7702, `extcodesize` will report > 0 code for the EOA addresses. So this would be allowed. It's rather odd to implement `IValidationModule` with an EOA account, but it would be possible.
2. The case we want to protect against is making the call to `IValidationModule.validateRuntime` to an address with zero code, because it will return success without actually executing anything. This is only a risk because `validateRuntime` has no return value, otherwise the ABI-decoding would fail when the address has no code. This is a check that Solidity inserts by default for all external calls without return data, so the assembly here is just copying what solidity already does. Info [here](https://alchemyinsights.slack.com/archives/C02VDQV5TKQ/p1710514180929839).

Zer0dot (approved):
Going ahead and approving for now, feel free to merge or wait for more eyes on this!

PR #226, opened 2024-10-06
feat: hook call buffers

adamegyed:
## Motivation

Much like the motivation for the user op validation and runtime validation call buffers, the account repeats in-memory encoding of external calls to perform pre-exec hooks. This can be avoided by reusing memory, in some cases even reusing memory previously used for runtime validation / pre runtime validation hooks.

## Solution

Implement this buffer reuse strategy.

Add a test that verifies all pre hook calls are still being made correctly, including strict ABI-encoding for the outgoing data.
(Does not yet have a fuzz test, only a manual unit test).

### Performance note

This increases the runtime gas costs of the no-hook cases by ~80 gas. This is purely from the new stack variables and conditional checks, and isn't from new memory usage. The gains in the >1 hook case should offset this, with ~2k gained in the "session key 1" case, which only has 1 pre exec hook. The next PR in the stack reduces this by more than this amount.


src/libraries/ExecutionLib.sol
```diff
@@ -461,6 +465,219 @@ library ExecutionLib {
...
+
+            // Need at least 64 bytes of return data to be considered successful.
+            success := and(success, gt(returndatasize(), 0x3f))
+            // Only accept return data of "strict encoding" form, where the relative offset is exactly 0x20.
+            success := and(success, eq(mload(0), 0x20))
+            // Ensure that the reported length of return data does not exceed the actual length.
+            // aka the stored length <= retundatasize() - 0x40 (for the first two values)
+            // No opcode for lte, so the expression equals:
+            // stored length < retundatasize() - 0x3f
+            // Underflow doesn't matter, because success is false anyways if length < 0x40.
```
Zer0dot, line 610: Can you elaborate on this? 

`call` will just not touch the remainder if the return data size is less than 0x40, why would success be false?

Then, there's the notion of dirty bytes-- since this is the scratch space, if the case I mentioned is true, it's possible that the length we'd be reading from `returnedBytesSize` is incorrect, right?

src/libraries/ExecutionLib.sol
```diff
@@ -461,6 +465,219 @@ library ExecutionLib {
...
+
+        assembly ("memory-safe") {
+            // Load the module address and entity Id
+            entityId := and(shr(64, hookEntity), 0xffffffff)
+            moduleAddress := shr(96, hookEntity)
+
+            // Update the buffer with the entity Id
+            mstore(add(buffer, 0x24), entityId)
+
+            // Perform the call, storing the first two words of return data into scratch space.
```
Zer0dot, line 584: Slick use of scratch space

src/account/ModularAccountBase.sol
```diff
@@ -282,23 +288,29 @@ abstract contract ModularAccountBase is
         ModuleEntity runtimeValidationFunction = ModuleEntity.wrap(bytes24(authorization[:24]));
 
         // Check if the runtime validation function is allowed to be called
-        bool isGlobalValidation = uint8(authorization[24]) == 1;
         _checkIfValidationAppliesCallData(
             data,
             runtimeValidationFunction,
-            isGlobalValidation ? ValidationCheckingType.GLOBAL : ValidationCheckingType.SELECTOR
+            // Unfortunately, have to avoid declaring a `bool isGlobalValidation` to avoid stack too deep issues.
```
Zer0dot, line 291: Made me realize how we're not treating this as validationFlags but in UO validation, we are. Nothing jumps out as incorrect, but worth noting.

src/libraries/ExecutionLib.sol
```diff
@@ -460,6 +464,219 @@ library ExecutionLib {
...
+
+            // Buffer transformation:
+            // 0xAAAAAAAA // selector
+            // 0x000: 0x________________________BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB account -> discarded
+            // 0x020: 0x________________________________________________________CCCCCCCC entityId -> selector
+            // 0x040: 0x________________________DDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD msg.sender -> entityId
+            // 0x060: 0xEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE msg.value -> sender
+            // 0x080: 0x______________________________________________________________c0 callData offset -> value
+            // 0x0a0: 0x_____________________________________________________________FFF auth offset -> cd offset
+            // 0x0c0: 0x_____________________________________________________________GGG callData length -> stays
```
Zer0dot, line 491: Correct me if I'm wrong but this property staying at the same offset from the rtCallBuffer is why we can use `getCallData` later in `executeWithRuntimeValidation()` with the rtCallBuffer which was already converted to a PHCallBuffer right?

adamegyed: Yes, the callData part stays intact.

test/account/PHCallBuffers.t.sol
```diff
@@ -0,0 +1,540 @@
...
+                    _hasPre: true,
+                    _hasPost: false
+                })
+            );
+        }
+
+        ValidationConfig validationConfig;
+
+        if (_isSMATest) {
+            // Only use the SMA for skipping the validation step, so install this with the fallback valitaion
```
Zer0dot:
```suggestion
            // Only use the SMA for skipping the validation step, so install this with the fallback validation
```

test/account/PHCallBuffers.t.sol
```diff
@@ -0,0 +1,540 @@
...
+    ModuleEntity internal _validationFunction;
+
+    event ReceivedCall(bytes msgData, uint256 msgValue);
+
+    function setUp() public override {
+        _allowTestDirectCalls();
+    }
+
+    // pre hooks in `executeUserOp` (freshly allocated buffer)
+    function test_preExecHooksRun_execUO() public {
```
Zer0dot: Think we're missing `withSMATest` in this file

test/account/DirectCallsFromModule.t.sol
```diff
@@ -123,6 +123,16 @@ contract DirectCallsFromModuleTest is AccountTestBase {
         account1.execute(CODELESS_ADDRESS, 0, "");
     }
 
+    function test_directCallFromModuleSequence_runHooks() external {
```
Zer0dot, line 126:
```suggestion
    function test_directCallFromModuleSequence_runHooks() external withSMATest {
```

Zer0dot (approved):
LGTM! Feel free to wait for another pair of eyes, or you have the go-ahead.

test/account/PHCallBuffers.t.sol
adamegyed: Yep, I now realize that's because these PRs were put up before we had the `withSMATest` modifier. Adding.

PR #227, opened 2024-10-07
feat: dense post hooks

adamegyed:
## Motivation

When saving pre hook return data to send to post hooks, our current setup needs to hold both the entire pre exec hooks array in memory, and all pre hook return datas in a `bytes[] memory`. These are consumed by the high-level call to IExecutionHookModule.postExecutionHook, which involves full memory copies of the data to send. We can improve performance by storing the post hook data in a format that is ready to be sent out in an external call, and skip the memory copy.

## Solution

Implement said optimization.

Add a fuzz test for the data sent to post exec hooks.

Define a new UDVT for the densely packed post hooks.


src/account/ModularAccountBase.sol
```diff
@@ -16,19 +15,34 @@ import {HookConfig, HookConfigLib} from "@erc6900/reference-implementation/libra
...
 import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import {IERC721Receiver} from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 import {MessageHashUtils} from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
+import {MessageHashUtils} from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
+import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import {UUPSUpgradeable} from "solady/utils/UUPSUpgradeable.sol";
+import {UUPSUpgradeable} from "solady/utils/UUPSUpgradeable.sol";
```
fangting-alchemy: All the imports are duplicates that should not be added. Might be from the merging conflict resolution.

adamegyed: Yep, it was from that. Fixed.

src/libraries/ExecutionLib.sol
```diff
@@ -677,6 +670,213 @@ library ExecutionLib {
...
+
+                // Perform the call
+                success :=
+                    call(
+                        gas(),
+                        moduleAddress,
+                        /*value*/
+                        0,
+                        /*argOffset*/
+                        workingMemPtr,
```
fangting-alchemy, line 800: Would this send over the current calldata and all the remaining hooks+data over?

adamegyed: No, it only sends the data from the corresponding pre hook, if there is one. Aka it matches the exact parameter format for `IExecutionHookModule.postExecutionHook`, which are `(uint32 entityId, bytes calldata preExecHookData)`, and doesn't send any extra data. This is asserted in the hooks test with:

```solidity
vm.expectEmit(address(execHooks[i]));
emit ReceivedCall(
    abi.encodeCall(
        IExecutionHookModule.postExecutionHook, (uint32(i), fuzzConfig.preHookReturnData[i])
    ),
    0
);
```

fangting-alchemy (approved):

PR #228, opened 2024-10-07
refactor: use reference-implementation's libraries

jaypaik:
- Removed `@erc6900/modular-account-libs` dependency as it is no longer being used.
- Updated `@erc6900/reference-implementation` dependency to v0.8.0-rc.4.
- Deleted HookConfigLib.sol and replaced with source from @erc6900/reference-implementation.
- Deleted ModuleEntityLib.sol and replaced with source from @erc6900/reference-implementation.
- Deleted ValidationConfigLib.sol and replaced with source from @erc6900/reference-implementation.
- Deleted ModuleEIP712.sol.
- Deleted ReplaySafeWrapper.sol and replaced with source from @erc6900/reference-implementation.

Will see if we want to update dedupe anything else after the other PRs are merged.


adamegyed (approved):
Looks good, thanks for handling!

PR #229, opened 2024-10-07
feat: merge allowlist and erc20 spend limit modules

fangting-alchemy:
## Motivation
ERC20Token spend limit cannot be used without Allowlist module.

## Solution
Merge ERC20 spend limit with allowlist modules.


src/modules/permissions/AllowlistModule.sol
```diff
@@ -25,22 +27,47 @@ import {BaseModule} from "../../modules/BaseModule.sol";
...
 ///       - if sepecific address + sepecific selector, pass
 ///       - revert all other cases
-contract AllowlistModule is IValidationHookModule, BaseModule {
+///    - Uninstallation will not remove all state for account. Account must pass all data to remove all rules for
+/// an validation entity.
+/// If specific address is set, a ERC20TokenSpendLimit might be set for it. If so, the following features and
+/// restrictions apply:
+///    - The execution hooks to be associated with validations should be used.
+///    - For any validation function with this execution hook installed, only token contracts with a set limit will
+/// be
```
Zer0dot: Maybe some minor formatting could be done here

Zer0dot: (I know how annoying the automatic line length adjustments from the formatter can be)

src/modules/permissions/AllowlistModule.sol
```diff
@@ -73,6 +105,29 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
...
+        if (selector == IModularAccount.execute.selector) {
+            // when calling execute or ERC20 functions directly
+            (address token,, bytes memory innerCalldata) = abi.decode(callData, (address, uint256, bytes));
+            _decrementLimitIfApplies(entityId, token, innerCalldata);
+        } else if (selector == IModularAccount.executeBatch.selector) {
+            Call[] memory calls = abi.decode(callData, (Call[]));
+            for (uint256 i = 0; i < calls.length; i++) {
+                _decrementLimitIfApplies(entityId, calls[i].target, calls[i].data);
+            }
+        } else {
```
Zer0dot, line 124: So basically, you should never install the validation-associated execution hook on anything but exec/execBatch right?

src/modules/permissions/AllowlistModule.sol
```diff
@@ -204,7 +280,36 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
...
+        uint256 spendLimit = erc20SpendLimits[entityId][token][msg.sender];
+
+        if (innerCalldata.length < 68) {
+            revert InvalidCalldataLength();
+        }
+
+        bytes4 selector;
+        uint256 spend;
+        assembly ("memory-safe") {
+            selector := mload(add(innerCalldata, 32)) // 0:32 is arr len, 32:36 is selector
```
Zer0dot:
```suggestion
            selector := mload(add(innerCalldata, 0x20)) // 0x00:0x20 is arr len, 0x20:0x24 is selector
```
Just for consistency when working with memory I think we opt to use hexadecimal offsets, but this is a nitpick.

src/modules/permissions/AllowlistModule.sol
```diff
@@ -204,7 +280,36 @@ contract AllowlistModule is IValidationHookModule, BaseModule {
...
+            spend := mload(add(innerCalldata, 68)) // 36:68 is recipient, 68:100 is spend
```
Zer0dot:
```suggestion
            spend := mload(add(innerCalldata, 0x44)) // 0x24:0x44 is recipient, 0x44:0x64 is spend
```
Same as above

test/modules/AllowlistERC20TokenLimit.t.sol
```diff
@@ -13,50 +13,64 @@ import {HookConfigLib} from "../../src/libraries/HookConfigLib.sol";
...
 import {ValidationConfigLib} from "../../src/libraries/ValidationConfigLib.sol";
-import {ERC20TokenLimitModule} from "../../src/modules/permissions/ERC20TokenLimitModule.sol";
+import {AllowlistModule} from "../../src/modules/permissions/AllowlistModule.sol";
 
 import {MockModule} from "../mocks/modules/MockModule.sol";
 import {AccountTestBase} from "../utils/AccountTestBase.sol";
 import {CODELESS_ADDRESS} from "../utils/TestConstants.sol";
 
-contract ERC20TokenLimitModuleTest is AccountTestBase {
+contract AllowlistERC20TokenLimitTest is AccountTestBase {
```
Zer0dot, line 23: Is there a reason we're not combining this test with the AllowlistModuleTest contract?

Also, can we add `withSMATest()` modifiers to the tests? (Except ones which only touch the module and not the account)

Zer0dot:
No further comments beyond what I've got, happy to approve once addressed!

src/modules/permissions/AllowlistModule.sol
fangting-alchemy: Depends on what you are trying to check on. In our case, `executeUserOp` calls self, so checking those two seems enough.

But users may install other execution functions that might do something. They can have a validation-associated hook on the validator that has access to that execution function.

test/modules/AllowlistERC20TokenLimit.t.sol
fangting-alchemy: I thought it was better organized this way to make sure we test both features of the module.

All tests have `withSMATest` modifier already.

src/modules/permissions/AllowlistModule.sol
Zer0dot: yeah but you shouldn't install *this* one on anything but those two or you will revert yeah?

fangting-alchemy: Sry, miss read. For this hook, yes.

Zer0dot (approved):

PR #231, opened 2024-10-07
refactor: codesize reduction

Zer0dot:
## Motivation

We need to reduce code size to fit under the Spurious Dragon limit of 24576 bytes.

## Solution

Various refactors, including:

- Merging certain similar functions and introducing branching.
- Removing `domainSeparator()` and potentially other offchain-computable functions.
- Move certain functions to external libraries if possible and the gas impact to do so is low.


src/account/ModularAccountBase.sol
```diff
@@ -332,7 +316,17 @@ abstract contract ModularAccountBase is
         ExecutionManifest calldata manifest,
         bytes calldata moduleUninstallData
     ) external override wrapNativeFunction {
-        _uninstallExecution(module, manifest, moduleUninstallData);
+        (bool success,) = _EXECUTION_INSTALL_DELEGATE.delegatecall(
```
adamegyed: Could we move the actual `delegatecall` step into the assembly block below? I think the solidity low-level call behavior is to collect the return data in memory even if it is discarded in the tuple.

src/account/ModularAccountBase.sol
```diff
@@ -322,7 +296,17 @@ abstract contract ModularAccountBase is
         ExecutionManifest calldata manifest,
         bytes calldata moduleInstallData
     ) external override wrapNativeFunction {
-        _installExecution(module, manifest, moduleInstallData);
+        (bool success,) = _EXECUTION_INSTALL_DELEGATE.delegatecall(
```
adamegyed: same here

src/account/ModularAccountBase.sol
Zer0dot: addressed in [563270](https://github.com/alchemyplatform/modular-account/pull/231/commits/d56327075cd3ec6422ffcd715144088aa28cb717)

src/account/ModularAccountBase.sol
Zer0dot: done in [563270](https://github.com/alchemyplatform/modular-account/pull/231/commits/d56327075cd3ec6422ffcd715144088aa28cb717)!

src/account/ModularAccountBase.sol
adamegyed: Did it end up changing gas usage or codesize? If not, then my info is outdated

Zer0dot: codesize looks to be the same, which I didn't expect. It's probably handled in the via-ir pipeline.

Zer0dot: Yup, saves codesize in the standard pipeline, not in `optimized-build`

adamegyed: Ah interesting, thanks for testing!

src/account/ModularAccountBase.sol
```diff
@@ -26,6 +26,8 @@ import {IERC165} from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import {UUPSUpgradeable} from "solady/utils/UUPSUpgradeable.sol";
 
 import {_coalescePreValidation, _coalesceValidation} from "../helpers/ValidationResHelpers.sol";
+
```
adamegyed: nit: no newline

src/libraries/ExecutionLib.sol
```diff
@@ -71,17 +71,56 @@ library ExecutionLib {
...
+                // we immediately revert, we can omit storing the length as we normally would for a `bytes memory`
+                // type, as well as omit finalizing the allocation by updating the free memory pointer.
+                let revertDataLocation := mload(0x40)
+                returndatacopy(revertDataLocation, 0, returndatasize())
+                revert(revertDataLocation, returndatasize())
+            }
+        }
+
+        // Manually call, without collecting return data unless there's a revert.
+        assembly ("memory-safe") {}
```
adamegyed: nit: discard

jaypaik (approved):

PR #233, opened 2024-10-08
feat: merge timeRange and paymaster guard modules

fangting-alchemy:


fangting-alchemy:
Merging makes it cheaper for users that use both hooks for the same validation. However, it makes it more expensive (installing and using) when the user only wants to use one of the two features for a validation.
This is an exploration and I don't think we should go this direction. Closing the PR given that.

PR #234, opened 2024-10-08
fix: deferred validation install encoding

adamegyed:
## Motivation

There are 2 issues we need to address for deferred validation install:
- The current abi-decoding logic for the struct `DeferredValidationInstallData` is vulnerable to signature extension, similar to [this issue](https://docs.safe.global/advanced/smart-account-bug-bounty/past-paid-bounties#signature-verification-does-not-enforce-a-maximum-size-on-the-signature-bytes) with Safe.
- The EIP-712 struct hash calculation for the type `InstallValidation` does not correctly encode dynamic fields.

## Solution

Rather than encoding and decoding a solidity struct (ABI tuple) for the parameters to `installValidation`, we can just encode and decode a fixed-length `bytes` call to `installValidation`. By doing so, the signature over the EIP-712 struct will sign over a fixed length in bytes, which would cause any extension attack on this field to result in a signature failure.

In the deferred install logic, assert that the first 4 bytes of the provided data are `installValidation`, and pull the `validationConfig` parameter out from the call to use for the rest of user op validation.

The new EIP-712 typed data struct is `InstallValidation(uint256 nonce,uint48 deadline,bytes validationInstall)`. Update the typed data hash calculation to compute the hash of dynamic fields with an additional `keccak256` operation.

Also update the various inner functions that aid in deferred validation install to not permanently allocate memory when it isn't needed to do so.

Because `domainSeparator` is deterministic on the account address and chainId, we don't need to expose a public function for it. Change it to internal-only to save a bit of codesize.

Update all deferred validation install tests and gas benchmarks.


src/account/ModularAccountBase.sol
```diff
@@ -931,6 +905,69 @@ abstract contract ModularAccountBase is
...
+            let installCallHash := keccak256(fmp, encodedInstallCall.length)
+
+            // Compute the struct hash
+            let ptr := fmp
+            mstore(ptr, _INSTALL_VALIDATION_TYPEHASH)
+            ptr := add(ptr, 0x20)
+            mstore(ptr, nonce)
+            ptr := add(ptr, 0x20)
+            // Clear the upper bits of the deadline, in case the caller didn't.
+            mstore(ptr, and(deadline, 0xffffffff))
```
Zer0dot, line 943: since the deadline passed appears to be a uint48 cast already, in which scenario is this possible?

adamegyed: In our usage, this will always be the case. But in general, for parameters of internal functions, it is not guaranteed that the input has upper bits cleared (when compiling with via-ir), so this is defensively masking it from potential bad callers. It's not strictly necessary.

Zer0dot: Pretty small anyway, fine to keep!

Zer0dot (approved):

PR #236, opened 2024-10-09
fix: module signature length checks

adamegyed:
## Motivation

To protect from a signature inflation attack, pre user op validation hooks should enforce an expected amount of per-validation-hook data.

## Solution

For our existing validation hook modules, assert the data length is zero.

Add tests asserting that passing data here causes a revert.

Update old tests to the new format, using the AccountTestBase harness and the `withSMATest` modifier.


test/modules/AllowlistModule.t.sol
```diff
@@ -1,13 +1,16 @@
...
-
 import {Call} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+import {MessageHashUtils} from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
 
 import {ModularAccountBase} from "../../src/account/ModularAccountBase.sol";
 import {HookConfigLib} from "../../src/libraries/HookConfigLib.sol";
 import {ModuleEntity, ModuleEntityLib} from "../../src/libraries/ModuleEntityLib.sol";
+
```
jaypaik: Nit: remove newline

test/modules/TimeRangeModule.t.sol
```diff
@@ -12,6 +12,8 @@ import {
 import {ModularAccountBase} from "../../src/account/ModularAccountBase.sol";
 import {HookConfigLib} from "../../src/libraries/HookConfigLib.sol";
 import {ModuleEntity, ModuleEntityLib} from "../../src/libraries/ModuleEntityLib.sol";
+
```
jaypaik: Nit: remove newline here too

jaypaik (approved):

jaypaik:
As part of my refactor and consolidation work, I'll probably port this change over to reference-implementation and import the module related contracts from there instead of having it exist in 2 places.

Edit:
Hmm seems like we may introduce more optimizations for BaseModule. Might just make sense to keep our copy.

test/modules/AllowlistModule.t.sol
adamegyed: argh the formatter strikes again

PR #237, opened 2024-10-11
feat: account code cleanup

adamegyed:
## Motivation

Account code cleanup.

## Solution

Each commit in this PR tracks a separate change, view them individually to get the specific context.

Changes:
- Benchmark batch calls
- Selector applicability checks decoding only in calldata
- Consolidate calldata slice revert conditions
- Skip return data collection on self call


src/account/ModularAccountBase.sol
```diff
@@ -838,28 +846,153 @@ abstract contract ModularAccountBase is
...
+            if gt(add(arrayPos, mul(callsLength, 0x20)), dataEnd) { revert(0, 0) }
+        }
+
+        // Now, we have the array length and data bounds.
+        // Iterate through the array elements, checking:
+        // - If the target is this account, assert that:
+        //   - the selector in the data field is not `execute` or `executeBatch`.
+        //   - the provided validation is allowed to call the selector.
+
+        for (uint256 i = 0; i < callsLength; ++i) {
```
fangting-alchemy, line 934: This is a lot complexity for the pretty simple for loop. Do we know how much are we saving for this the change (did not see any of the setup in gas does executeBatch)?

adamegyed: It saves ~1.7k in the batch action (batch size 2, for transferring both native token and ERC-20). Yeah agree it is complex 😬 , but it's basically just a copy-paste from compiler output.

src/account/ModularAccountBase.sol
```diff
@@ -838,28 +846,153 @@ abstract contract ModularAccountBase is
...
+            // Set up the "safe data decoding range"
+            let headStart := callData.offset
+            dataEnd := add(headStart, callData.length)
+
+            // Assert it is safe to load the offset
+            if slt(sub(dataEnd, headStart), 32) { revert(0, 0) }
+
+            // Load and sanitize the offset
+            let relOffset := calldataload(callData.offset)
+            if gt(relOffset, 0xffffffffffffffff) { revert(0, 0) }
```
Zer0dot, line 911: What's the point of this check exactly?

src/account/ModularAccountBase.sol
```diff
@@ -838,28 +846,153 @@ abstract contract ModularAccountBase is
...
+        // The length of the `Call[]` array.
+        uint256 callsLength;
+
+        assembly ("memory-safe") {
+            // Set up the "safe data decoding range"
+            let headStart := callData.offset
+            dataEnd := add(headStart, callData.length)
+
+            // Assert it is safe to load the offset
+            if slt(sub(dataEnd, headStart), 32) { revert(0, 0) }
```
Zer0dot, line 907: Can you elaborate on how this slt ensures it's safe to load the offset?

src/account/ModularAccountBase.sol
Zer0dot: If it's copied from the compiler why is it different from what the original code compiles to?

fangting-alchemy: Is the 1.7k savings all from the for loop rewrite?
Each of those calls should not burn more than 200 gas. What did I miss?

src/account/ModularAccountBase.sol
```diff
@@ -838,28 +846,153 @@ abstract contract ModularAccountBase is
...
+
+            // Convert from a relative offset to an absolute offset.
+            let absOffset := add(headStart, relOffset)
+
+            // Assert it is safe to load the length
+            if iszero(slt(add(absOffset, 0x1f), dataEnd)) { revert(0, 0) }
+
+            // Load and sanitize the length
+            callsLength := calldataload(absOffset)
+            if gt(callsLength, 0xffffffffffffffff) { revert(0, 0) }
```
Zer0dot, line 921: same q re: sanitizing, not sure I follow how the length is only invalid if it's greater than this-- is there a more particular reason for this constant?

src/account/ModularAccountBase.sol
fangting-alchemy: My understanding is that This is probably to ensure the actual calldata is as long as 32 bytes (the size of the offset). If it is equal, it means 0 calls passed.

src/account/ModularAccountBase.sol
fangting-alchemy: Confused by this check as well.

src/account/ModularAccountBase.sol
adamegyed: Yep, mirroring what @fangting-alchemy said. It asserts that the position to load from (via `calldataload`) is still within the bounds of the `bytes calldata` type, including all 32 bytes that `calldataload` gets at once.

src/account/ModularAccountBase.sol
adamegyed: It is asserting that the offset value is not "too big", where "too big" is determined by the compiler. This prevents any overflow issues (from the unchecked addition ops), and is a bit stricter in total size.

For context while reviewing this entire section, these checks aren't things that I came up with, they're just a copy-paste of compiler output for decoding data into a type of `Call[] calldata`. So, to have the behavior exactly match between validation (here) and execution, I ported the exact same logic over.

src/account/ModularAccountBase.sol
adamegyed: It's the default solidity compiler behavior, mirror here again.

src/account/ModularAccountBase.sol
adamegyed: Because this is copied from the compiler output decoding a parameter of type `Call[] calldata`. Currently, the solidity compiler only supports doing this decoding at the top of a function (aka how you get variables of that type to begin with), but it doesn't support doing this within the body of a function via `abi.decode(...)`. It only supports decoding from `memory` to `memory`, while this logic is decoding from `calldata` to `calldata`.

src/account/ModularAccountBase.sol
```diff
@@ -838,28 +846,153 @@ abstract contract ModularAccountBase is
...
-                        // Note that modules of other contracts using `executeWithRuntimeValidation` may still
-                        // independently call into this account with a different validation function, allowing
-                        // composition of multiple batches.
-                        revert SelfCallRecursionDepthExceeded();
+        // The absolute offset of the start of the `Call[]` array.
+        uint256 arrayPos;
+        // The length of the `Call[]` array.
+        uint256 callsLength;
+
+        assembly ("memory-safe") {
```
fangting-alchemy, line 901:
```suggestion
        // This block is retrieving the actual Call[] location, length, the end of the calldata.
        assembly ("memory-safe") {
```

fangting-alchemy (approved):
Looks good. Please add more docs as discussed in the sync.

PR #239, opened 2024-10-11
feat: change deferred validation to deferred action

adamegyed:
## Motivation

To support token paymasters that need approvals to happen before their validation, we can extend the functionality of deferred validation install to perform any arbitrary action, provided the signer has permission to do so.

## Solution

Rename "deferred validation install" to "deferred action", and update the EIP-712 struct name.

Previously, we pulled the validation function to use out from the parameter to `installValidation`. Given that this call is now able to perform arbitrary actions, we must instead define a separate field in the EIP-712 struct for the validation to use after the deferred action, for user op validation. Also update all existing tests for deferred validation install to use this new field.

The added calldata of the above change slightly increases the gas costs of using deferred validation install.

Add a test for the ERC-20 approve during account deployment validation, including a mock token paymaster.

Also add a test for the correct denial of deferred actions that the caller lacks permissions for.


src/account/ModularAccountBase.sol
```diff
@@ -453,30 +446,30 @@ abstract contract ModularAccountBase is
             bytes calldata encodedData = userOp.signature[29:29 + encodedDataLength];
 
             // Get the deferred installation signature length.
```
Zer0dot:
```suggestion
            // Get the deferred action signature length.
```

src/account/ModularAccountBase.sol
```diff
@@ -432,19 +431,13 @@ abstract contract ModularAccountBase is
...
+        if (hasDeferredAction) {
             // Use outer validation as a 1271 validation, then use the installed validation to validate the rest.
-            // Check if the outer validation applies to `installValidation`.
-            _checkIfValidationAppliesSelector(
-                this.installValidation.selector,
-                validationFunction, // Treated as sig val
-                isGlobalValidation ? ValidationCheckingType.GLOBAL : ValidationCheckingType.SELECTOR
-            );
 
             // Get the length of the abi-encoded `DeferredValidationInstallData` struct.
```
Zer0dot:
```suggestion
            // Get the length of the deferred action data.
```

src/account/ModularAccountBase.sol
```diff
@@ -432,19 +431,13 @@ abstract contract ModularAccountBase is
         /// Byte indices are [inclusive, exclusive]
         ///      [25:29] : uint32, encodedDatalength.
         ///      [29:(29 + encodedDatalength)] : bytes, abi-encoded deferred validation data.
```
Zer0dot:
```suggestion
        ///      [29:(29 + encodedDatalength)] : bytes, abi-encoded deferred action data.
```

src/account/ModularAccountBase.sol
```diff
@@ -432,19 +431,13 @@ abstract contract ModularAccountBase is
...
-        ///      [(29 + encodedDataLength):(33 + encodedDataLength)] : uint32, deferredInstallSigLength.
-        ///      [(33 + encodedDataLength):(33 + deferredInstallSigLength + encodedDataLength)] : bytes,
+        ///      [(29 + encodedDataLength):(33 + encodedDataLength)] : uint32, deferredActionSigLength.
+        ///      [(33 + encodedDataLength):(33 + deferredActionSigLength + encodedDataLength)] : bytes,
         ///         deferred install sig. This is the signature passed to the outer validation decoded earlier.
```
Zer0dot:
```suggestion
        ///         deferred action sig. This is the signature passed to the outer validation decoded earlier.
```

src/account/ModularAccountBase.sol
```diff
@@ -432,19 +431,13 @@ abstract contract ModularAccountBase is
...
-        ///      [(33 + deferredInstallSigLength + encodedDataLength):] : bytes, userOpSignature. This is the
+        ///      [(33 + deferredActionSigLength + encodedDataLength):] : bytes, userOpSignature. This is the
         ///         signature passed to the newly installed deferred validation.
-        if (isDeferredInstallValidation) {
+        if (hasDeferredAction) {
             // Use outer validation as a 1271 validation, then use the installed validation to validate the rest.
```
Zer0dot:
```suggestion
            // Use outer validation as a 1271 validation, then use the inner validation to validate the UO.
```

src/account/ModularAccountBase.sol
```diff
@@ -432,19 +431,13 @@ abstract contract ModularAccountBase is
...
         ///      [25:29] : uint32, encodedDatalength.
         ///      [29:(29 + encodedDatalength)] : bytes, abi-encoded deferred validation data.
-        ///      [(29 + encodedDataLength):(33 + encodedDataLength)] : uint32, deferredInstallSigLength.
-        ///      [(33 + encodedDataLength):(33 + deferredInstallSigLength + encodedDataLength)] : bytes,
+        ///      [(29 + encodedDataLength):(33 + encodedDataLength)] : uint32, deferredActionSigLength.
+        ///      [(33 + encodedDataLength):(33 + deferredActionSigLength + encodedDataLength)] : bytes,
         ///         deferred install sig. This is the signature passed to the outer validation decoded earlier.
-        ///      [(33 + deferredInstallSigLength + encodedDataLength):] : bytes, userOpSignature. This is the
+        ///      [(33 + deferredActionSigLength + encodedDataLength):] : bytes, userOpSignature. This is the
         ///         signature passed to the newly installed deferred validation.
```
Zer0dot:
```suggestion
        ///         signature passed to the inner validation.
```

src/account/ModularAccountBase.sol
```diff
@@ -453,30 +446,30 @@ abstract contract ModularAccountBase is
...
             validationData = uint256(deadline) << 160;
 
             // Call `installValidation` on the account.
-            ExecutionLib.callBubbleOnRevertTransient(address(this), 0, encodedData[38:]);
-
-            // Update the UserOp signature to the remaining bytes.
-            userOpSignature = userOp.signature[33 + encodedDataLength + deferredInstallSigLength:];
+            ExecutionLib.callBubbleOnRevertTransient(address(this), 0, encodedData[63:]);
 
             // Load in the validation that was just installed.
```
Zer0dot:
```suggestion
            // Load in the inner validation.
```

src/account/ModularAccountBase.sol
```diff
@@ -423,7 +422,7 @@ abstract contract ModularAccountBase is
         // Decode the 25th byte into an 8-bit bitmap (6 bits of which remain unused).
         uint8 validationFlags = uint8(userOp.signature[24]);
         bool isGlobalValidation = validationFlags & _IS_GLOBAL_VALIDATION_BIT != 0;
-        bool isDeferredInstallValidation = validationFlags & _IS_DEFERRED_INSTALL_VALIDATION_BIT != 0;
+        bool hasDeferredAction = validationFlags & _HAS_DEFERRED_ACTION_BIT != 0;
 
         // Assigned depending on whether the UO uses deferred validation installation or not.
```
Zer0dot:
```suggestion
        // Assigned depending on whether the UO includes a deferred action or not.
```

src/account/ModularAccountBase.sol
```diff
@@ -453,30 +446,30 @@ abstract contract ModularAccountBase is
...
             // Get the deferred installation signature length.
-            uint256 deferredInstallSigLength =
+            uint256 deferredActionSigLength =
                 uint32(bytes4(userOp.signature[29 + encodedDataLength:33 + encodedDataLength]));
 
+            // Update the UserOp signature to the remaining bytes.
+            userOpSignature = userOp.signature[33 + encodedDataLength + deferredActionSigLength:];
+
             // Get the deferred installation signature, which is passed to the outer validation to install the
             // deferred validation.
```
Zer0dot:
```suggestion
            // deferred action.
```

src/account/ModularAccountBase.sol
```diff
@@ -453,30 +446,30 @@ abstract contract ModularAccountBase is
...
 
             // Get the deferred installation signature length.
-            uint256 deferredInstallSigLength =
+            uint256 deferredActionSigLength =
                 uint32(bytes4(userOp.signature[29 + encodedDataLength:33 + encodedDataLength]));
 
+            // Update the UserOp signature to the remaining bytes.
+            userOpSignature = userOp.signature[33 + encodedDataLength + deferredActionSigLength:];
+
             // Get the deferred installation signature, which is passed to the outer validation to install the
```
Zer0dot:
```suggestion
            // Get the deferred installation signature, which is passed to the outer validation to handle the
```

src/account/ModularAccountBase.sol
```diff
@@ -509,42 +502,48 @@ abstract contract ModularAccountBase is
         }
     }
 
-    /// @return The deadline of the deferred validation.
-    function _validateDeferredInstallDataAndSetNonce(
+    /// @return The deadline of the deferred validation and the validation function to use.
```
Zer0dot:
```suggestion
    /// @return The deadline of the deferred action and the validation function to use.
```

src/account/ModularAccountBase.sol
```diff
@@ -509,42 +502,48 @@ abstract contract ModularAccountBase is
...
+    function _validateDeferredActionAndSetNonce(
         ModuleEntity sigValidation,
+        bool isGlobalValidation,
         bytes calldata encodedData,
         bytes calldata sig
-    ) internal returns (uint48) {
+    ) internal returns (uint48, ValidationConfig) {
         // Decode stack vars for the deadline and nonce.
```
Zer0dot, line 512:
```suggestion
        // Decode stack vars for the deadline and nonce.
        // The deadline, nonce, inner validation, and deferred call selector are all at fixed positions in the encodedData.
```

test/account/DeferredAction.t.sol
```diff
@@ -0,0 +1,172 @@
...
+
+        vm.prank(beneficiary);
+        entryPoint.handleOps(userOps, beneficiary);
+
+        assertEq(10 ether, erc20.balanceOf(address(paymaster)));
+    }
+
+    // Install a new execution function that the validation does not have the privilege to call.
+    // Assert that user op validation reverts if the deferred action tries to call it.
+    function test_noPrivilegeEscalation_deferredAction() public {
```
Zer0dot:
```suggestion
    function test_deferredAction_noPrivilegeEscalation() public {
```
Smol nit

test/account/DeferredAction.t.sol
```diff
@@ -0,0 +1,172 @@
...
+            ValidationConfigLib.pack(_signerValidation, true, false, false),
+            account1,
+            owner1Key,
+            uoSig
+        );
+
+        PackedUserOperation[] memory userOps = new PackedUserOperation[](1);
+        userOps[0] = userOp;
+
+        vm.prank(beneficiary);
```
Zer0dot, line 161: Sorry why do we prank the beneficiary here?

Zer0dot (approved):

test/account/DeferredAction.t.sol
adamegyed: Just out of habit, because it's how a bundler would usually work. In theory this doesn't change anything, but sometimes with things like `_allowTestDirectCalls()`, the test contract itself can get some privileges wrt/ the account. But this is the EP, so it shouldn't matter.

Zer0dot:
Note: we just need to pass over test functions that still refer to deferred validation installation

PR #240, opened 2024-10-12
feat: signature call buffers

adamegyed:
## Motivation

The last external call that the account makes without call buffers is signature validation, and signature validation hooks.

## Solution

Implement call buffers for signature validation.

Add a fuzz test for signature validation inputs.

Reset memory after signature validation in deferred actions to reuse the memory during execution.


src/account/ModularAccountBase.sol
```diff
@@ -541,9 +540,12 @@ abstract contract ModularAccountBase is
             uoValidation
         );
 
+        // Clear the memory after performing signature validation
+        MemSnapshot memSnapshot = MemManagementLib.freezeFMP();
```
Zer0dot, line 544: dude this feels like a sick pattern we could probably use pretty often, but is it safe? Are we sure we're not making obsolete some memory we might not know about?

adamegyed: I couldn't really think of another place to use it: it doesn't matter for UO validation, and for RT validation, we only allocate the call buffer after loading the pre-validation hooks, and that buffer may be reused for pre exec hooks, so we can't easily free the memory used to hold the pre RT validation hooks.

In theory, the only memory this operation makes unsafe are blocks newly allocated in between the freeze and the restore. Within the internal function `_isValidSignature`, we allocate an array for the signature validation hooks, and a call buffer for signature validation. Afaict neither of these are reused, nor make it out of the internal function `_isValidSignature`, so I think we should be fine to deallocate them and reuse that space for the UO validation hooks + UO call buffer..

src/libraries/ExecutionLib.sol
```diff
@@ -811,6 +813,161 @@ library ExecutionLib {
...
+        assembly ("memory-safe") {
+            // Load the module address and entity id
+            let entityId := and(shr(64, hookEntity), 0xffffffff)
+            let moduleAddress := shr(96, hookEntity)
+
+            // Update the buffer with the entity Id
+            mstore(add(buffer, 0x44), entityId)
+
+            // Copy in the signature segment
+            // Since the buffer's copy of the signature exceeds the length of any sub-segments, we can safely write
```
Zer0dot, line 856: Actually @adamegyed is it possible to fuck this up with some sneaky encoding from the user?

Zer0dot: i.e. I encode a length longer than the entire sig for a single segment

adamegyed: Good thinking! This worried me at first, but I think we're ok.

This should be caught by the behavior of the calldata slices: https://docs.soliditylang.org/en/v0.8.26/types.html#array-slices
> If start is greater than end or if end is greater than the length of the array, an exception is thrown.

### Looking at our code:
We pass in to SparseCalldataSegmentLib a bounded `bytes calldata` type, which is either `userOp.signature[25:]`, the version that skips past the deferred action data, `bytes calldata authorization`, or `bytes calldata signature`.

In SparseCalldataSegmentLib:
Loading the length:
https://github.com/alchemyplatform/modular-account/blob/6313e6ccb1ec98acb09452230374f11b69cb64ad/src/libraries/SparseCalldataSegmentLib.sol#L63-L64
Getting the new slice:
https://github.com/alchemyplatform/modular-account/blob/6313e6ccb1ec98acb09452230374f11b69cb64ad/src/libraries/SparseCalldataSegmentLib.sol#L38

So if the reported length exceeds the actual length, the slice creation step should revert.

src/libraries/ExecutionLib.sol
```diff
@@ -811,6 +813,161 @@ library ExecutionLib {
...
+                    actualCallLength,
+                    /*retOffset*/
+                    0,
+                    /*retSize*/
+                    0x20
+                )
+
+            if iszero(success) {
+                // Bubble up the revert if the call reverts.
+                let m := mload(0x40)
```
Zer0dot, line 892: What's the reasoning behind calling the fmp "m?"

src/account/ModularAccountBase.sol
```diff
@@ -753,45 +755,45 @@ abstract contract ModularAccountBase is
...
+        bytes32 hash,
+        ModuleEntity sigValidation,
+        bytes calldata signatureSegment
+    ) internal view virtual returns (bytes4) {
+        (hash); // unused
         AccountStorage storage _storage = getAccountStorage();
 
-        (address module, uint32 entityId) = sigValidation.unpack();
         if (!_storage.validationData[sigValidation].isSignatureValidation) {
+            (address module, uint32 entityId) = sigValidation.unpack();
```
Zer0dot, line 792: small idea but can we just emit the module entity without unpacking, what are the implications? (I know we do this in other places too, just wondering if we could save a handful of codesize)

src/libraries/ExecutionLib.sol
```diff
@@ -893,4 +1050,18 @@ library ExecutionLib {
             mstore(authorizationOffsetPtr, sub(authorizationOffset, 0x20))
         }
     }
+
+    function _prepareSigValidationCallBufferPreSigValidationHooks(SigCallBuffer buffer) private pure {
+        uint32 selector = uint32(IValidationHookModule.preSignatureValidationHook.selector);
+        assembly ("memory-safe") {
+            // Update the buffer with the selector. This will squash a portion of the `account` param for runtime
```
```suggestion
            // Update the buffer with the selector. This will squash a portion of the `account` param for signature
```

src/libraries/ExecutionLib.sol
adamegyed: I think it's just "memory", like "memory we can use". iirc this was a convention that solady was doing for a bit, but I can see how it's not great for readability. What about "mem" or "freeMem"?

src/account/ModularAccountBase.sol
```diff
@@ -753,45 +755,45 @@ abstract contract ModularAccountBase is
...
-        virtual
-        returns (bytes4)
-    {
+    function _exec1271Validation(
+        SigCallBuffer buffer,
+        bytes32 hash,
+        ModuleEntity sigValidation,
+        bytes calldata signatureSegment
+    ) internal view virtual returns (bytes4) {
+        (hash); // unused
```
Zer0dot: what's this for?

adamegyed: SMA needs access to the hash, and doesn't need a call buffer. If we were to fully make the MA and SMA code independent, we could omit this.

adamegyed: Should probably document in a comment, adding now.

src/libraries/ExecutionLib.sol
```diff
@@ -893,4 +1050,18 @@ library ExecutionLib {
...
+            // validation, but that will be restored before calling.
+            mstore(add(buffer, 0x24), selector)
```
Zer0dot, line 1059: Is it possible to omit the cast to uint32 at the start of this function, and just `mstore(add(buffer, 0x40), selector)` instead? Since both in sig validation and pre sig validation, we override the memory at 0x44?

(Edit: 0x40 cause we need to push by 28 bytes since the bytes4 is left-aligned)

src/account/ModularAccountBase.sol
adamegyed: Probably, we should go back and review our error types to make sure the data included is correct. Like here, I agree we should keep it packed.

src/libraries/ExecutionLib.sol
adamegyed: I mostly did this to save a minute amount of codesize - the `bytes4` immediate value will have to be left-aligned, so it will always be a `push32`, vs the `uint32` can be a `push4`.

test/account/SigCallBuffer.t.sol
```diff
@@ -0,0 +1,215 @@
...
+    function _installValidationAndAssocHook(FuzzConfig memory fuzzConfig) internal {
+        HookConfig[] memory hooks = new HookConfig[](fuzzConfig.validationHookCount);
+        bytes[] memory hookInstalls = new bytes[](fuzzConfig.validationHookCount);
+
+        ExecutionManifest memory m; // empty manifest
+        validationHooks = new MockModule[](fuzzConfig.validationHookCount);
+
+        for (uint256 i = 0; i < fuzzConfig.validationHookCount; i++) {
+            // To get different addresses for vm.expectCall between SMA and non-SMA tests, we need to deploy new
+            // the validation hooks at different addresses
```
Zer0dot:
```suggestion
            // validation hooks at different addresses
```

Zer0dot: smart way to work around the expectCall limitation

test/account/SigCallBuffer.t.sol
```diff
@@ -0,0 +1,215 @@
...
+    function testFuzz_sigCallBuffer(bytes32 hash, FuzzConfig memory fuzzConfig) public withSMATest {
+        _installValidationAndAssocHook(fuzzConfig);
+
+        _expectCalls(fuzzConfig, hash);
+
+        PreValidationHookData[] memory preValidationHookDatasToSend = _generatePreHooksDatasArray(fuzzConfig);
+
+        bytes memory signature;
+        if (_isSMATest) {
+            // Signature will fail, but not revert.
```
Zer0dot, line 112: Can you elaborate?

adamegyed: It will return `0xffffffff`, but all of the hooks will still run and nothing will revert. Updating the comment now.

Zer0dot (approved):
Minor comments, lgtm!

src/libraries/ExecutionLib.sol
Zer0dot: oh scratch that my mental model was wrong, this is strictly better

src/account/ModularAccountBase.sol
Zer0dot: Either way nbd, I don't think it's going to make much difference, but devX is nicer if we don't have to do extra decoding upon receiving an error. Fine either way.

src/libraries/ExecutionLib.sol
Zer0dot: either way works with me. Happy to keep it as is, it's informational.

src/libraries/ExecutionLib.sol
Zer0dot: Right, good stuff! This means the segment *could* be incorrect in that it could overreach into other segments (fine, no danger afaik), but it cannot under any circumstance break the invariant of being less than the total signature length. Neato.

src/account/ModularAccountBase.sol
Zer0dot: Makes sense, my concern is if Solc adds some memory allocations in between. I think it's fine if our tests pass, though.

src/libraries/ExecutionLib.sol
adamegyed: Actually, I think if it "overreaches into the other segments", the other segments would fail to load, because the index and length is only read from after the length of the segment. So the definition of the format makes that kind of overlap impossible.

PR #241, opened 2024-10-12
feat: add ERC7739 compatibility

howydev:
Switch the current `ReplaySafeHash` scheme to use an ERC7739 compatible signature scheme instead

Implementation uses pieces from, and/or is inspired by Solady and ERC7579's ERC7739 validator implementation


src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // !bytes32 salt;
+        // !uint256[] memory extensions;
+        /// @solidity memory-safe-assembly
+        assembly {
+            //Rebuild domain separator out of 712 domain
+            let m := mload(0x40) // Load the free memory pointer.
+            mstore(m, _DOMAIN_SEPARATOR_TYPEHASH_ACCOUNT)
+            mstore(add(m, 0x20), _HASH_EMPTY_BYTES) // Name hash.
```
howydev: this smells wrong.. think we should either not have this, or also be hashing `salt` and `extensions`

howydev: maybe its ok, since its used for PersonalSign, whereas for TypedDataSign we use all fields even if its empty

src/modules/validation/WebauthnValidationModule.sol
```diff
@@ -23,7 +24,7 @@ import {BaseModule} from "@erc6900/reference-implementation/modules/BaseModule.s
 ///    - This validation supports ERC-1271. The signature is valid if it is signed by the owner's private key.
 ///    - This validation supports composition that other validation can relay on entities in this validation
 /// to validate partially or fully.
-contract WebauthnValidationModule is IValidationModule, ReplaySafeWrapper, BaseModule {
+contract WebauthnValidationModule is IValidationModule, ERC7739ReplaySafeWrapper, BaseModule {
```
adamegyed: We should be OK to omit this, right? Since no webauthn signer will be handled by an `eth_signTypedData_v4` or `personal_sign`, so we can continue to use the ReplaySafeWrapper.

test/modules/WebauthnValidationModule.t.sol
```diff
@@ -45,16 +46,40 @@ contract WebauthnValidationModuleTest is AccountTestBase {
...
+            ModularAccountBase(account).isValidSignature(
+                mockAppDigest, _get1271SigForChallenge(challenge, mockAppStructHash, 0, 0)
+            ) == 0x1626ba7e
         );
     }
 
-    function testFuzz_fail_isValidSignature(bytes32 message, uint256 sigR, uint256 sigS) external view {
-        bytes32 challenge = module.replaySafeHash(account, message);
+    // Fuzz sig
+    function test_fail_isValidSignature(bytes32 message, uint256 sigR, uint256 sigS) external view {
```
adamegyed: Should start with `testFuzz`

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+        assembly {
+            m := mload(0x40) // Grab the free memory pointer.
+            mstore(0x40, add(m, 0x120)) // Allocate the memory.
+            // Skip 2 words for the `typedDataSignTypehash` and `contents` struct hash.
+            mstore(add(m, 0x40), shl(248, fields)) // TODO: maybe parse fields and only use the necessary ones
+            mstore(add(m, 0x60), _HASH_EMPTY_BYTES)
+            mstore(add(m, 0x80), _HASH_EMPTY_BYTES)
+            mstore(add(m, 0xa0), chainid())
+            mstore(add(m, 0xc0), account)
+            // mstore(add(m, 0xe0), address(0))
```
Zer0dot: What's this? Looks like we're missing a word here?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // bytes32 salt = bytes32(uint160(account));
+        // !uint256[] memory extensions;
+        /// @solidity memory-safe-assembly
+        assembly {
+            m := mload(0x40) // Grab the free memory pointer.
+            mstore(0x40, add(m, 0x120)) // Allocate the memory.
+            // Skip 2 words for the `typedDataSignTypehash` and `contents` struct hash.
+            mstore(add(m, 0x40), shl(248, fields)) // TODO: maybe parse fields and only use the necessary ones
```
Zer0dot: TODO remains here, which of the two options is the canonical ERC721 way of doing things? (i.e. is it correct how we do it or should we parse the fields?)

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    function _typedDataSignFieldsForAccount(address account) internal view virtual returns (bytes32 m) {
+        bytes1 fields = bytes1(hex"0C"); // 001100 or 001110
+        // !string memory name;
+        // !string memory version;
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // bytes32 salt = bytes32(uint160(account));
+        // !uint256[] memory extensions;
+        /// @solidity memory-safe-assembly
+        assembly {
```
Zer0dot:
```suggestion
        assembly ("memory-safe") {
```
Idk if this makes a difference, but IIRC this is the new way of doing things. Smol nit though

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L253
+    function _typedDataSignFieldsForAccount(address account) internal view virtual returns (bytes32 m) {
+        bytes1 fields = bytes1(hex"0C"); // 001100 or 001110
+        // !string memory name;
+        // !string memory version;
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // bytes32 salt = bytes32(uint160(account));
+        // !uint256[] memory extensions;
+        /// @solidity memory-safe-assembly
```
Zer0dot:
```suggestion
```

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+        assembly {
+            m := mload(0x40) // Grab the free memory pointer.
+            mstore(0x40, add(m, 0x120)) // Allocate the memory.
+            // Skip 2 words for the `typedDataSignTypehash` and `contents` struct hash.
+            mstore(add(m, 0x40), shl(248, fields)) // TODO: maybe parse fields and only use the necessary ones
+            mstore(add(m, 0x60), _HASH_EMPTY_BYTES)
+            mstore(add(m, 0x80), _HASH_EMPTY_BYTES)
+            mstore(add(m, 0xa0), chainid())
+            mstore(add(m, 0xc0), account)
```
Zer0dot: just to make sure I'm understanding correctly, this is only the case since the verifyingContract, address(this), is the same as account for this case?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+            mstore(add(m, 0x80), account)
+            digest := keccak256(m, 0xa0) //domain separator
+
+            // Hash typed data
+            mstore(0x00, 0x1901000000000000) // Store "\x19\x01".
+            mstore(0x1a, digest) // Store the domain separator.
+            mstore(0x3a, structHash) // Store the struct hash.
+            digest := keccak256(0x18, 0x42)
+            // Restore the part of the free memory slot that was overwritten.
+            mstore(0x3a, 0)
```
Zer0dot: couldn't this overwrite the free memory pointer by replacing with zero?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+            let m := mload(0x40) // Load the free memory pointer.
+            mstore(m, _DOMAIN_SEPARATOR_TYPEHASH_ACCOUNT)
+            mstore(add(m, 0x20), _HASH_EMPTY_BYTES) // Name hash.
+            mstore(add(m, 0x40), _HASH_EMPTY_BYTES) // Version hash.
+            mstore(add(m, 0x60), chainid())
+            mstore(add(m, 0x80), account)
+            digest := keccak256(m, 0xa0) //domain separator
+
+            // Hash typed data
+            mstore(0x00, 0x1901000000000000) // Store "\x19\x01".
```
Zer0dot: why all the zeroes at the end?

Zer0dot: I guess to offset it to the left, but in that case why not offset all the way and have it be at the start of the word, and start the digest at 0x02?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+            mstore(0x1a, digest) // Store the domain separator.
+            mstore(0x3a, structHash) // Store the struct hash.
+            digest := keccak256(0x18, 0x42)
+            // Restore the part of the free memory slot that was overwritten.
+            mstore(0x3a, 0)
+        }
+    }
+
+    // Used in the module
```
Zer0dot: Do you think it's worth separating into two contracts, one for modules and one for SMA? 

Not a big deal, just somewhat fearful this would leave some artifacts in the SMA

Zer0dot: Actually why not make this a lib?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+            mstore(add(m, 0xa0), account)
+            digest := keccak256(m, 0xa0) //domain separator
+
+            // Hash typed data
+            mstore(0x00, 0x1901000000000000) // Store "\x19\x01".
+            mstore(0x1a, digest) // Store the domain separator.
+            mstore(0x3a, structHash) // Store the struct hash.
+            digest := keccak256(0x18, 0x42)
+            // Restore the part of the free memory slot that was overwritten.
+            mstore(0x3a, 0)
```
Zer0dot: same q about overwriting the fmp

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @dev `keccak256("PersonalSign(bytes prefixed)")`.
+    bytes32 internal constant _PERSONAL_SIGN_TYPEHASH =
+        0x983e65e5148e570cd828ead231ee759a8d7958721a768f93bc4483ba005c32de;
+
+    // The difference between a module hash and an account hash is:
+    // Account domains include only chainId and verifyingContract of itself (not the implementation)
+    // Module domains include chainId, verifyingContract of the module, and uses the optional salt param, using the
+    // account address
+
+    // keccak256("EIP712Domain(uint256 chainId,address verifyingContract, bytes32 salt)")
```
adamegyed: This comment is wrong, the space would change the hash output. Thankfully the actual hex value is correct.
```suggestion
    // keccak256("EIP712Domain(uint256 chainId,address verifyingContract,bytes32 salt)")
```

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+// A contract mixin ERC7739-compliant nested EIP-712 wrappers over EIP-1271 digests.
+// This allows for efficient, readable ERC-1271 signature schemes for smart contract accounts.
+contract ERC7739ReplaySafeWrapper {
```
adamegyed: If this is only a mixin, this should be `abstract contract`.

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+        return (hash, signature);
+    }
+
+    /// @notice Helper function to obtain and arrange the eip712 fields for an account
+    /// @dev From
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L283
+    /// which is based on
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L253
+    function _typedDataSignFieldsForAccount(address account) internal view virtual returns (bytes32 m) {
+        bytes1 fields = bytes1(hex"0C"); // 001100 or 001110
```
adamegyed:
```suggestion
        bytes1 fields = bytes1(hex"0C"); // 001100
```
This is just the former. I might be missing something, but it doesn't look like it ever takes the value of 0b001110 (0x0E).

src/helpers/ERC7739ReplaySafeWrapper.sol
adamegyed: I think it is assuming that the upper 6 bytes of the FMP will never not be zero, because that would be a huge amount of memory to allocate.

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @notice Helper function to obtain and arrange the eip712 fields for an account
+    /// @dev From
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L283
+    /// which is based on
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L253
+    function _typedDataSignFieldsForModule(address account, address module)
+        internal
+        view
+        virtual
+        returns (bytes32 m)
```
adamegyed: By our current conventions, we should define a custom type instead of using `bytes32` here.

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: but 3a + 0x20 is 0x5a, so we're zeroing out the leftmost 26 bytes, leaving only 6 bytes not zeroed, I guess that's also technically okay but seems a little more risky.

adamegyed: Oh shoot, yeah you're right. I agree it feels kinda sketchy.

Zer0dot: Still technically 2^48 bytes of memory to use, but I still am not sure. Would rather just restore a cached fmp, wdyt?

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: good catch!

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev:
```suggestion
abstract contract ERC7739ReplaySafeWrapper {
```

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: we're not using this field, so the field will be 0, but cheaper to do nothing than to mstore 0

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: unfortunately we use this helper in our tests too :/

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: Hmm 6 bytes of memory sounds risky, but 2^48 is ~10e14? That seems reasonably safe. Happy to change it, or leave a note and call it out for our auditors

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: yeah it probably should be a lib. I think I had state in it at one point

howydev: On separation - either way is probably fine, but I lean towards keeping it in the same file since if we split, `_validateERC7739SigFormat` will be duplicated across both. Stylistically feels weird to dup that since that function is the main function in the lib

(This is assuming the compiler would ignore internal functions that aren't used + simplify the jumps if there's only 1 jump origin and 1 jump destination, but I could be wrong there..)

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: Maybe a nit but memory past the FMP is not necessarily zeroed out, so it may be dirty. C.C. @adamegyed for confirmation

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: Not sure I follow, is it always hardcoded in the account?

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: I'm not sure if it's worth doing this over just restoring the fmp normally, probably a little bit of extra gas but it saves us the potential issues with this. My vote is for not messing with this and restoring the fmp normally, but it should be fine either way.

src/helpers/ERC7739ReplaySafeWrapper.sol
fangting-alchemy: +1 we never use version.

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,259 @@
...
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L283
+    /// which is based on
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L253
+    function _typedDataSignFieldsForAccount(address account) internal view virtual returns (bytes32 m) {
+        bytes1 fields = bytes1(hex"0C"); // 001100 or 001110
+        // !string memory name;
+        // !string memory version;
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // bytes32 salt = bytes32(uint160(account));
```
fangting-alchemy: This should be updated. I believe salt is also not used.

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,259 @@
...
+            mstore(add(m, 0xc0), account)
+            // mstore(add(m, 0xe0), address(0))
+            mstore(add(m, 0x100), _HASH_EMPTY_BYTES)
+        }
+    }
+
+    /// @notice Helper function to build the full personal sign hash
+    /// @dev From
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L312
+    function _hashTypedDataAccount(address account, bytes32 structHash)
```
fangting-alchemy: bit
```suggestion
    function _hashTypedDataForAccount(address account, bytes32 structHash)
```

src/helpers/ERC7739ReplaySafeWrapper.sol
fangting-alchemy: Hmm, should we use `_HASH_EMPTY_BYTES` is we are not using the field like name and version?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,259 @@
...
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // bytes32 salt = address(account);
+        // !uint256[] memory extensions;
+        /// @solidity memory-safe-assembly
+        assembly {
+            //Rebuild domain separator out of 712 domain
+            let m := mload(0x40) // Load the free memory pointer.
+            mstore(m, _DOMAIN_SEPARATOR_TYPEHASH_MODULE)
+            mstore(add(m, 0x20), _HASH_EMPTY_BYTES) // Name hash.
```
Zer0dot: Why do we need to include these empty bytes if per eip712 we only need to include the relevant fields in the domain separator struct-- right?

Zer0dot: Actually are we sure the domain separator typehash works here, cause this looks like it should include name and version, which we just treat as empty, rather than not include them at all

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: good catch, that was from an earlier version before i split up `_typedDataSignFields` into 2 separate functions

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: good catch

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: good question, not fully sure. my guess is that each 0 byte thats added here ends up in bytecode, and 6 bytes is the intersection of "secure enough" and "maximally efficient" from a gas + bytecode perspective to Vectorized

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: I think I just realized what you mean.. 

Your suggestion is to remove the 6 bytes from `x19x01`, save the domain/hashStruct in 0x20 and 0x40 respectively, and fully restore the fmp after? Not sure if im missing something, but it does seem like an improvement

Also another possibility is to overwrite memory at the current fmp.. We already use memory from `m` to `add(m, 0ax0)` so theres no additional expansion, and we wouldn't need to do any restoration after

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: its just syntactic sugar right? do we incur any overhead with doing it in a type?

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: this confused me for a while also, but got some clarification after discussing with @adamegyed yesterday

These fields come from 712, but whether to include them into the struct is a 7739 convention. Because the account type we're using is a fixed: `TypedDataSign(XXXX contents, contents,bytes1 fields,string name,string version,uint256 chainId,address verifyingContract,bytes32 salt,uint256[] extensions)`, this forces the client has to include all fields even if they're blank, else the user's signature would be over the wrong typehash -> the wrong domain

I think Adam also suggested 7739 supporting variability in what fields to include - the verifier would build a differently sized `_hashTypedDataModule` based on the `fields` bitmap from 5267, but then the implementation would have to be dynamically build the full type based on the same `fields` bitmap

howydev: Its an interesting suggestion and quite possibly a cleaner/better one. Due to time constraints we have to stay with this one though

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: Ah thought i removed this, must have lost that change somewhere.. dived into this a little in a thread below

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: oh whoops, total brain fart, read your prev comment wrongly. Yep, verifyingContract is the account for the SMA

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: Unfortunately I think we have to use `_HASH_EMPTY_BYTES`.. by 712 convention, to encode any non 32 byte type, the field is hashed. Based on this, my understanding is that if the client does `empty string ""` in the field, signTypedData would use `HASH_EMPTY_BYTES` to build the digest that the signature is over

This is one of the things im not 100% certain about and think we should e2e test it just to make sure

src/modules/validation/WebauthnValidationModule.sol
howydev: Hmm I guess if its necessary, since 7739 is implemented in the module that flow could be supported by webauthn 7739 module

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: on second pass, think this is ok

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: But then shouldn't the domain separator also include those empty fields?

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: In this case maybe we can shorten it for SMA and have the function just call address() instead of taking a param wdyt?

howydev: do you mean doing `address()` here, or in SMA?

Zer0dot: in SMA with a different function is my thought

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: What if we just mstore(0x40, m)?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @param hash The hash to wrap.
+    /// @return bytes32 Replay-safe hash, computed by wrapping the input hash in an EIP-712 struct.
+    /// @return bytes Inner signature to use for verification.
+    function _validateERC7739SigFormat(bytes32 t, bytes32 hash, bytes calldata signature)
+        private
+        pure
+        returns (bytes32, bytes32, bytes calldata)
+    {
+        /// @solidity memory-safe-assembly
+        assembly {
```
Zer0dot: @howydev this asm is a little hard to read, I think it's better to keep the comments on separate lines from the asm, and space out asm lines that aren't really related, and smol nit but maybe bigger variable names instead of single letters

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// replay
+    /// @dev Parses out the inner signature from the full signature and returns it
+    /// @dev Implementation is lifted from
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L191
+    /// @dev Also see:
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L22
+    /// @dev Only change is: based on the SignatureType, we perform either ECDSA verification or ERC-1271
+    /// verification.
+    /// @param t Type to use
+    /// @param hash The hash to wrap.
```
jaypaik: Mind making these more descriptive?

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: Sorry read the comment, this is the same as Solady's right?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    /// @notice Helper function to validate ERC7739 compatible nested EIP712 structs to guard against signature
+    /// replay
+    /// @dev Parses out the inner signature from the full signature and returns it
+    /// @dev Implementation is lifted from
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L191
+    /// @dev Also see:
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L22
+    /// @dev Only change is: based on the SignatureType, we perform either ECDSA verification or ERC-1271
```
jaypaik: This isn't obvious in the function body - this happens outside of this right? Maybe clarify this and mention the actual difference between `_erc1271IsValidSignatureViaNestedEIP712` and this function.

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: possible, so we would have 3 diff `typedDataSignFieldsForAccount` type of functions. The gains are minimal so will keep this as an idea for the future

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: the main app domain could, but it doesnt need to because its has flexibility on defining its own domain type (the keccak256("EIP712Domain(string name,string version,uint256 chainId)), but we need to because 7739 currently specifies that all fields are used

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: Yup, good catch, will update

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: yup this is solady's. I think I can add higher level explainers for what each block is trying to do, do you think that would help?

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: > What if we just mstore(0x40, m)?  

that's a solution thats relatively safer than the current one (`mstore(3a, 0)` being relatively less safe), but by the gas bad principle i think we should just use the space after fmp since it [saves 3 gas](https://github.com/alchemyplatform/modular-account/pull/241/commits/aa4f20f800fefc390541d420e7de4ebf53e3852e) from not having to do any memory restoration. Unsure about bytecode size between the 2 though, do you know how we can quickly test and compare that?

src/modules/validation/WebauthnValidationModule.sol
howydev: ok, reverted to the prev ReplaySafeWrapper for webauthn

test/account/ModularAccount.t.sol
```diff
@@ -47,6 +47,7 @@ contract ModularAccountTest is AccountTestBase {
     event ReceivedCall(bytes msgData, uint256 msgValue);
 
     function setUp() public override {
+        _revertSnapshot = vm.snapshot();
```
howydev, line 51: more context on why this is here - in the repo, to do tests on MA + SMA we take a snapshot in the test constructor in `AccountTestBase` and this saves a copy of the chain state. The `withSMATest` modifier causes tests to run once on MAs, then does a revert to the snapshotted state and runs tests on SMAs

However, because it happens during the constructor, when the state is resetted, the test contract ends up having 0 bytecode. For some reason this causes external calls to revert which we rely on here as an easy way to convert `bytes memory` into `bytes calldata` 

This PR moves the snapshot into `setUp()`, so when reverting to the snapshot, the test contract address will contain the bytecode with the required external helper functions

test/utils/ModuleSignatureUtils.sol
```diff
@@ -28,17 +29,90 @@ contract ModuleSignatureUtils {
     bytes32 private constant _INSTALL_VALIDATION_TYPEHASH =
         keccak256("InstallValidation(uint256 nonce,uint48 deadline,bytes validationInstall)");
 
+    // 712 for a Mock App
+    string internal constant _MOCK_APP_CONTENTS_TYPE = "Message(string message)"; // len 23
+    bytes32 internal constant _MOCK_APP_DOMAIN = 0x71062c282d40422f744945d587dbf4ecfd4f9cfad1d35d62c944373009d96162;
+
+    string internal constant _DEFERRED_INSTALL_CONTENTS_TYPE =
+        "InstallValidation(uint256 nonce,uint48 deadline,bytes validationInstall)"; // len 36
```
adamegyed: Heads up, this type has changed in `develop` to `"DeferredAction(uint256 nonce,uint48 deadline,bytes25 validationFunction,bytes call)"`

test/utils/ModuleSignatureUtils.sol
```diff
@@ -28,17 +29,90 @@ contract ModuleSignatureUtils {
...
+        returns (bytes32 mockAppStructHash, bytes32 mockAppDigest)
+    {
+        mockAppStructHash = keccak256(abi.encode(keccak256(abi.encodePacked(_MOCK_APP_CONTENTS_TYPE)), _digest));
+        assembly {
+            mstore(0x00, 0x1901000000000000)
+            mstore(0x1a, _MOCK_APP_DOMAIN)
+            mstore(0x3a, mockAppStructHash)
+            mockAppDigest := keccak256(0x18, 0x42)
+            mstore(0x3a, 0)
+        }
```
adamegyed: For a test-only file, could we avoid using assembly to be slightly less error-prone?
```suggestion
        mockAppDigest = keccak256(abi.encodePacked(bytes2(hex"19_01"), _MOCK_APP_DOMAIN, mockAppStructHash));
```

test/account/PerHookData.t.sol
```diff
@@ -400,20 +401,35 @@ contract PerHookDataTest is CustomValidationTestBase {
...
+        );
 
-        bytes32 replaySafeHash = ecdsaValidationModule.replaySafeHash(address(account1), messageHash);
         (uint8 v, bytes32 r, bytes32 s) = vm.sign(owner1Key, replaySafeHash);
 
         PreValidationHookData[] memory preValidationHookData = new PreValidationHookData[](1);
-        preValidationHookData[0] = PreValidationHookData({index: 0, validationData: abi.encodePacked(message)});
+        preValidationHookData[0] = PreValidationHookData({
+            index: 0,
+            validationData: abi.encodePacked(hex"1901", _MOCK_APP_DOMAIN, mockAppStructHash) // preimage of
```
adamegyed, line 422: Thanks for patching this!

test/utils/ModuleSignatureUtils.sol
```diff
@@ -7,14 +7,15 @@ import {MessageHashUtils} from "@openzeppelin/contracts/utils/cryptography/Messa
 import {Vm} from "forge-std/src/Vm.sol";
 
 import {ModularAccount} from "../../src/account/ModularAccount.sol";
-import {SemiModularAccountBytecode} from "../../src/account/SemiModularAccountBytecode.sol";
 import {RESERVED_VALIDATION_DATA_INDEX} from "../../src/helpers/Constants.sol";
-import {ECDSAValidationModule} from "../../src/modules/validation/ECDSAValidationModule.sol";
+
```
adamegyed: nit: remove newline

src/account/SemiModularAccountBase.sol
```diff
@@ -15,9 +14,12 @@ import {RTCallBuffer, UOCallBuffer} from "../libraries/ExecutionLib.sol";
 import {SemiModularKnownSelectorsLib} from "../libraries/SemiModularKnownSelectorsLib.sol";
 import {ModularAccountBase} from "./ModularAccountBase.sol";
```
adamegyed, line 18: nit: no newline here

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+// A contract mixin ERC7739-compliant nested EIP-712 wrappers over EIP-1271 digests.
```
adamegyed: super tiny nit, now that we've refactored this to be a library instead:
```suggestion
// A library for ERC7739-compliant nested EIP-712 wrappers over EIP-1271 digests.
```
We should also move the file to `src/libraries/`.

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,278 @@
...
+
+    // keccak256("EIP712Domain(uint256 chainId,address verifyingContract,bytes32 salt)")
+    bytes32 internal constant _DOMAIN_SEPARATOR_TYPEHASH_MODULE =
+        0x71062c282d40422f744945d587dbf4ecfd4f9cfad1d35d62c944373009d96162;
+
+    // keccak256("EIP712Domain(uint256 chainId,address verifyingContract)")
+    bytes32 internal constant _DOMAIN_SEPARATOR_TYPEHASH_ACCOUNT =
+        0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
+
+    // keccak256(abi.encode(""))
```
adamegyed: Comment nit: this will actually result in a hash of 
```
0x
0000000000000000000000000000000000000000000000000000000000000020 // offset
0000000000000000000000000000000000000000000000000000000000000000 // length
```
Which has a different hash.
This should instead say:
```suggestion
    // keccak256("")
```

Alternatively, we could switch it out for abi.encodePacked, but I think the shorter one makes sense here.

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,278 @@
...
+        // !string memory version;
+        // uint256 chainId = chainId;
+        // address verifyingContract = address(this);
+        // !bytes32 salt;
+        // !uint256[] memory extensions;
+        assembly ("memory-safe") {
+            m := mload(0x40) // Grab the free memory pointer.
+            mstore(0x40, add(m, 0x120)) // Allocate the memory.
+            // Skip 2 words for the `typedDataSignTypehash` and `contents` struct hash.
+            mstore(add(m, 0x40), shl(248, fields))
```
adamegyed: Wait... fields is already a `bytes1` type, which is left-aligned. Wouldn't this left shift make the entire contents zero?

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,278 @@
...
+        // uint256 chainId = chainId;
+        // address verifyingContract = module;
+        // bytes32 salt = account;
+        // !uint256[] memory extensions;
+        /// @solidity memory-safe-assembly
+        assembly {
+            m := mload(0x40) // Grab the free memory pointer.
+            mstore(0x40, add(m, 0x120)) // Allocate the memory.
+            // Skip 2 words for the `typedDataSignTypehash` and `contents` struct hash.
+            mstore(add(m, 0x40), shl(248, fields))
```
adamegyed: Same here, I think the expression `shl(248, fields)` will always be zero.

From testing in chisel:
```solidity
➜ bytes1 fields = bytes1(hex"0E");
➜ fields << 248
Type: uint256
├ Hex: 0x0
├ Hex (full word): 0x0
└ Decimal: 0
```

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+// A contract mixin ERC7739-compliant nested EIP-712 wrappers over EIP-1271 digests.
+// This allows for efficient, readable ERC-1271 signature schemes for smart contract accounts.
+library ERC7739ReplaySafeWrapper {
+    type TypeStruct is bytes32;
```
adamegyed:
```suggestion
        // Points to a location in memory with EIP-712 formatted `encodeType(TypedDataSign)`, excluding the first two words for typeHash(TypedDataSign)` and `typeHash(contents)`. Does not store the length, because this has a fixed size of `0x120` (9 words).
        // Used to compute `hashStruct(typedDataSign)`
        type TypeStruct is bytes32;
```
Might need to re-run the formatter after this.

adamegyed (commented):
Mostly looks good, but we should investigate the `fields` left shift issue.

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: added some explainers

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: lgtm as is, bytecode size won't be significantly affected for sure. How I check is I just:
```
git checkout HEAD~1
FOUNDRY_PROFILE=optimized-build forge build --sizes
(note the result)
git switch -
FOUNDRY_PROFILE=optimized-build forge build --sizes
```
And just compare, definitely a better way to do this but that's what I do lol

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: looks like it'd be correct if we just remove the shl, or shl a uint8

src/helpers/ERC7739ReplaySafeWrapper.sol
Zer0dot: lgtm, no stress (is Solady audited?)

src/helpers/ERC7739ReplaySafeWrapper.sol
adamegyed: I think we should just omit it. It looks like this issue is not present in solady, but they do have an operation that just cleans the lower 31 bytes (They do `shl(248, byte(0, fields)`, which is equivalent to just `fields` if the var is already clean).

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: good catch!

src/helpers/ERC7739ReplaySafeWrapper.sol
howydev: good catch! not 100% sure about which one the 712 would use for an empty string, but think its more likely to be `keccak256("")`. will make the change

test/utils/AccountTestBase.sol
```diff
@@ -280,46 +280,48 @@ abstract contract AccountTestBase is OptimizedTest, ModuleSignatureUtils {
...
-                    account,
-                    deferredInstallNonce,
-                    deferredInstallDeadline,
-                    uoValidationFunction,
-                    deferredValidationInstallCall
-                ),
-                signingKey,
+        bytes memory deferredValidationSig;
+        bytes memory deferredValidationDatas;
+        {
```
howydev, line 285: stack too deep

src/helpers/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,282 @@
...
+    /// @notice Helper function to obtain and arrange the eip712 fields for a module
+    /// @dev From
+    /// github/erc7579/erc7739Validator/blob/f8cbd4b58a7226cce18e9b8bc380da51174daf53/src/ERC7739Validator.sol#L283
+    /// which is based on
+    /// github/Vectorized/solady/blob/351548a824d57c1c0fec688fdfe3a44a8e17efc3/src/accounts/ERC1271.sol#L253
+    /// @param account The account address
+    /// @param module The module address
+    /// @return m The location of all the types
+    function typedDataSignFieldsForModule(address account, address module) internal view returns (TypeStruct m) {
+        bytes1 fields = bytes1(hex"0E"); // 001110
```
adamegyed: I think we want this to be 
```suggestion
        bytes1 fields = bytes1(hex"1C"); // 011100
```

Because salt is the last param, and the EIP-5267 spec says that the bitmap goes from LSB to MSB.

src/libraries/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,278 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+// A library for ERC7739-compliant nested EIP-712 wrappers over EIP-1271 digests.
+// This allows for efficient, readable ERC-1271 signature schemes for smart contract accounts.
+library ERC7739ReplaySafeWrapper {
+    type TypeStruct is bytes32;
```
adamegyed: Nit: type comment from previous review: 
```suggestion
    // Points to a location in memory with EIP-712 formatted `encodeType(TypedDataSign)`, excluding the first two words for typeHash(TypedDataSign)` and `typeHash(contents)`. Does not store the length, because this has a fixed size of `0x120` (9 words).
    type TypeStruct is bytes32;
```

Might need to re-run formatter.

src/libraries/ERC7739ReplaySafeWrapper.sol
```diff
@@ -0,0 +1,281 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+// A library for ERC7739-compliant nested EIP-712 wrappers over EIP-1271 digests.
+// This allows for efficient, readable ERC-1271 signature schemes for smart contract accounts.
+library ERC7739ReplaySafeWrapper {
```
jaypaik: Perhaps rename to ERC7739Lib or ERC7739ReplaySafeWrapperLib? Can follow up in another PR though.

adamegyed (approved):

PR #243, opened 2024-10-14
fix: correctly check hooks length in installValidation

adamegyed:
## Motivation

We don't actually ever revert with the custom error for exceeding the max hook count because the check addition would overflow before that. The custom error is preferred for visibility into the issue.

## Solution

Correctly detect exceeding the length and revert.


src/account/ModuleManagerInternals.sol
```diff
@@ -43,7 +43,7 @@ abstract contract ModuleManagerInternals is IModularAccount {
     error ExecutionHookAlreadySet(HookConfig hookConfig);
     error ModuleInstallCallbackFailed(address module, bytes revertReason);
     error ModuleNotInstalled(address module);
-    error PreValidationHookLimitExceeded();
+    error ValidationAssocHookLimitExceeded();
```
jaypaik: Nit: sort

jaypaik (commented):
What's the reason we don't need to enforce this for (non-validation associated) execution hooks again?

adamegyed:
> What's the reason we don't need to enforce this for (non-validation associated) execution hooks again?

We don't have a length field tracking it anywhere, so we don't know how long the existing set of hooks is when we're adding to it.

jaypaik (approved):

PR #246, opened 2024-10-15
fix: accountstorage coverage, natspec, storage root derivation

adamegyed:
## Motivation

Cleanup pass on `AccountStorage.sol` and `AccountStorageInitializable.sol`.

## Solution

Expand and clarify natspec.

Test an edge case for the disableInitializers logic (impossible to reach in MA/SMA code).

Update the derivation of the storage root. Previously, we were not actually using ERC-7201, and the hash calculation was wrong. Both of those are now corrected, and the derivation matches the pattern used in MAv1, but with the major version number increased to 2.

### DIFF
--- a/src/account/AccountStorage.sol
+++ b/src/account/AccountStorage.sol
@@ -5,17 +5,18 @@ import {HookConfig, ModuleEntity} from "@erc6900/reference-implementation/interf
 
 import {LinkedListSet, SetValue} from "../libraries/LinkedListSetLib.sol";
 
-// bytes = keccak256("ERC6900.ModularAccount.Storage")
-bytes32 constant _ACCOUNT_STORAGE_SLOT = 0xc531f081ecdb5a90f38c197521797881a6e5c752a7d451780f325a95f8b91f45;
+// ERC-7201 derived storage slot.
+// keccak256(abi.encode(uint256(keccak256("Alchemy.ModularAccount.Storage_V2")) - 1)) & ~bytes32(uint256(0xff))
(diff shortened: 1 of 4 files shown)


src/account/AccountStorage.sol
```diff
@@ -5,17 +5,18 @@ import {HookConfig, ModuleEntity} from "@erc6900/reference-implementation/interf
 
 import {LinkedListSet, SetValue} from "../libraries/LinkedListSetLib.sol";
 
-// bytes = keccak256("ERC6900.ModularAccount.Storage")
-bytes32 constant _ACCOUNT_STORAGE_SLOT = 0xc531f081ecdb5a90f38c197521797881a6e5c752a7d451780f325a95f8b91f45;
+// ERC-7201 derived storage slot.
+// keccak256(abi.encode(uint256(keccak256("Alchemy.ModularAccount.Storage_V2")) - 1)) & ~bytes32(uint256(0xff))
+bytes32 constant _ACCOUNT_STORAGE_SLOT = 0x596912a710dec01bac203cb0ed2c7e56a2ce6b2a68276967fff6dd57561bdd00;
 
 // Represents data associated with a specifc function selector.
```
jaypaik: Should we natspec this? I think foundry picks it up

jaypaik (approved):

src/account/AccountStorage.sol
adamegyed: Yep, updating

jaypaik: Oh I meant this line (also a typo)
```suggestion
/// @dev Represents data associated with a specific function selector.
```

adamegyed: Fixed!

PR #255, opened 2024-10-16
fix: correctly report hooks for account native functions

adamegyed:
## Motivation

We don't correctly show selector-associated execution hooks for account native function in the view function `getExecutionData`.

## Solution

Move the hook retrieval logic to the shared code branch.

Switch to using the KnownSelectorsLib for comparing for equality.

### DIFF
--- a/src/account/ModularAccountView.sol
+++ b/src/account/ModularAccountView.sol
@@ -1,43 +1,34 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.26;
 
-import {UUPSUpgradeable} from "@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol";
-
-import {
-    HookConfig,
-    IModularAccount,
-    ModuleEntity
-} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
+import {HookConfig, ModuleEntity} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
 import {
     ExecutionDataView,
     IModularAccountView,
     ValidationDataView
 } from "@erc6900/reference-implementation/interfaces/IModularAccountView.sol";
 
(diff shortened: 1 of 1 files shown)


src/account/ModularAccountView.sol
```diff
@@ -1,43 +1,34 @@
...
     function getExecutionData(bytes4 selector) external view override returns (ExecutionDataView memory data) {
-        if (
-            selector == IModularAccount.execute.selector || selector == IModularAccount.executeBatch.selector
-                || selector == UUPSUpgradeable.upgradeToAndCall.selector
-                || selector == IModularAccount.installExecution.selector
-                || selector == IModularAccount.uninstallExecution.selector
-        ) {
+        ExecutionStorage storage executionStorage = getAccountStorage().executionStorage[selector];
+
+        if (KnownSelectorsLib.isNativeFunction(selector)) {
```
fangting-alchemy, line 20: This func is gonna be retired soon. We will need a different way to decide if it's a native function.

adamegyed: Hmm... if so, I think we may be blocked and can't retire it. The spec says that this is now native functions should be reported in the view functions.

I'm pretty sure the if/else check is the only reliable way to test if it is a native function, do you know any other mechanism?

fangting-alchemy: hmm, i am not sure if there is another reliable way.
Even if we don't retire it in the lib, it won't be able to catch both SMA and MA cases with how the code resize refactor PR is set up. We will have to somehow allow overriding.

fangting-alchemy: We could split the other func checks out and keep the native one in as how it is in our current codebase. cc @Zer0dot

Zer0dot (approved):

fangting-alchemy (approved):

PR #257, opened 2024-10-16
fix: ModularAccountBase cleanup, coverage, custom error types

adamegyed:
## Motivation

Cleanup pass on `ModularAccountBase.sol`.

During this pass, discovered that we lack custom errors for UO and Sig validation module functions.

## Solution

- Add a script to get sizes
- Expand test coverage for MABase
- Remove duplicate error definitions
- Simplify error definitions in MABase
- Add custom errors for all module functions in ExecutionLib.


src/account/ModularAccountBase.sol
```diff
@@ -87,15 +87,11 @@ abstract contract ModularAccountBase is
...
-    error PreRuntimeValidationHookFailed(address module, uint32 entityId, bytes revertReason);
     error RequireUserOperationContext();
-    error RuntimeValidationFunctionReverted(address module, uint32 entityId, bytes revertReason);
     error SelfCallRecursionDepthExceeded();
-    error SignatureValidationInvalid(address module, uint32 entityId);
-    error UserOpValidationInvalid(address module, uint32 entityId);
-    error UnexpectedAggregator(address module, uint32 entityId, address aggregator);
+    error SignatureValidationInvalid(ModuleEntity validationFunction);
+    error UserOpValidationInvalid(ModuleEntity validationFunction);
+    error UnexpectedAggregator(ModuleEntity validationFunction, address aggregator);
```
jaypaik, line 101: Why are these`ModuleEntity` while the execution lib errors below are unpacked?

Also nit: Mind sorting this entire block of errors? (see CreateFailed at the bottom)

adamegyed: Last I tested, this was both cheaper and smaller in codesize, but let me re-test to be sure.

adamegyed: Ok it is actually ever so slightly less codesize - updated!

src/libraries/ExecutionLib.sol
```diff
@@ -196,35 +203,51 @@ library ExecutionLib {
...
                 validationData := mload(0)
             }
+        } else {
+            // Revert with the appropriate error type for the selector used.
+
+            uint32 selectorUsed;
+            uint32 errorSelector;
+
+            assembly ("memory-safe") {
+                selectorUsed := and(mload(add(buffer, 0x4)), 0xffffffff)
```
Zer0dot, line 241: clever way to load the first 4 bytes of 0x20!

jaypaik (approved):

src/libraries/ExecutionLib.sol
```diff
@@ -1029,6 +1057,37 @@ library ExecutionLib {
         return workingMemPtr;
     }
 
+    function _revertModuleFunction(uint32 errorSelector, address moduleAddress, uint32 entityId) private pure {
+        // All of the module function reverts have the same parameter layout:
+        // - module address
+        // - entity Id
+        // - revert data
+
+        assembly ("memory-safe") {
```
Zer0dot, line 1066: Is it possible to do 
```
revert(abi.encodeWithselector(this.error.selector, ...)
```

If so, are we sure it'd be cheaper to do this?

Zer0dot (approved):
small comment, lgtm

src/libraries/ExecutionLib.sol
adamegyed: If we do that, then we'd have to call the high-level function `collectRevertData` first, and the `abi.encode...` expression would copy that memory over to the new memory with the data for the revert. I believe this saves on both gas and codesize due to that added copy.

Zer0dot: gotcha, lgtm!

PR #258, opened 2024-10-16
chore: add integration tests for paymasterGuardModule

fangting-alchemy:
Add integration tests for paymaster guard module

### DIFF
--- a/test/modules/PaymasterGuardModule.t.sol
+++ b/test/modules/PaymasterGuardModule.t.sol
@@ -2,42 +2,67 @@
 pragma solidity ^0.8.26;
 
 import {PackedUserOperation} from "@eth-infinitism/account-abstraction/interfaces/PackedUserOperation.sol";
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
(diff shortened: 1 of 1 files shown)


test/modules/PaymasterGuardModule.t.sol
```diff
@@ -2,63 +2,90 @@
...
-
 import {AccountTestBase} from "../utils/AccountTestBase.sol";
 
 contract PaymasterGuardModuleTest is AccountTestBase {
+    using ECDSA for bytes32;
+    using MessageHashUtils for bytes32;
+
     PaymasterGuardModule public module = new PaymasterGuardModule();
 
     address public account;
```
adamegyed, line 23: Quick question: if we're using `account1`, why do we define `account` here?

adamegyed (approved):
Looks good, had 1 small question.

test/modules/PaymasterGuardModule.t.sol
fangting-alchemy: Unit tests used account and used a direct `onInstall` call without going through the MA. 
Don't have time to update all the tests. So using comments to separate the two tests types.

PR #263, opened 2024-10-16
fix: accountId and moduleId cleanup

jaypaik:


src/account/ModularAccount.sol
```diff
@@ -13,6 +15,11 @@ import {ModularAccountBase} from "./ModularAccountBase.sol";
 contract ModularAccount is ModularAccountBase {
     constructor(IEntryPoint anEntryPoint) ModularAccountBase(anEntryPoint) {}
 
+    /// @inheritdoc IModularAccount
+    function accountId() external pure virtual returns (string memory) {
+        return "alchemy.modular-account.2.0.0";
```
jaypaik: This is 2.0.0 since it's technically our second modular account. The SMAs are 1.0.0.

adamegyed (approved):

PR #265, opened 2024-10-16
docs: add natspec in libraries

adamegyed:
## Motivation

More thorough documentation.

## Solution

Cleanup comments and expand natspec in libraries.


src/libraries/MemManagementLib.sol
```diff
@@ -150,6 +174,8 @@ library MemManagementLib {
         return target;
     }
 
+    /// @notice Captures a snapshot of the free memory pointer.
+    /// @return snapshot The snapshot of the free memory pointer.
```
howydev:
```suggestion
    /// @return MemSnapshot The snapshot of the free memory pointer.
```

howydev (approved):

PR #271, opened 2024-10-24
chore: update forge-std and update deprecated cheatcodes

adamegyed:
## Motivation

We are currently getting warnings during test runs when using `vm.snapshot` and `vm.revertTo`, as these are now deprecated and replaced with `vm.snapshotState` and `vm.revertToState`.

## Solution

Swap the deprecated cheatcodes for their non-deprecated replacements.

~~Update forge-std to `master` and replace usage of the deprecated opcodes. `forge-std`'s latest release `v1.9.3` does not yet support the non-deprecated cheatcodes.~~

Update: since then, forge-std v1.9.4 has been released with this included in the changes. The PR has been updated to use this tagged version, rather than an arbitrary commit.


howydev (approved):
LGTM

PR #278, opened 2024-11-07
fix: [CL ALCHEMY-001] validation applicability check in deferred action

adamegyed:
## Motivation

Addresses CL ALCHEMY-001.

In `_validateDeferredActionAndSetNonce`, the incorrect validation applicability checking function was used, which only verified the selector, rather than considering the allowed pattern of batching using self-calls.

Note that we can't adopt the suggested remediation due to the requirement of supporting calls to `approve` on token contracts during a user op with initcode. See this test file for a sample case: https://github.com/alchemyplatform/modular-account/blob/develop/test/account/DeferredAction.t.sol#L61.

## Solution

Update the deferred action checking logic to use `_checkIfValidationAppliesCallData`, which performs the necessary checks over self-calls.

Also add a test verifying that the validation applicability check for self-calls is enforced.

### DIFF
--- a/gas-snapshots/ModularAccount.json
+++ b/gas-snapshots/ModularAccount.json
@@ -12,5 +12,5 @@
   "UserOp_NativeTransfer": "161531",
   "UserOp_UseSessionKey_Case1_Counter": "195236",
   "UserOp_UseSessionKey_Case1_Token": "226220",
-  "UserOp_deferredValidation": "257078"
+  "UserOp_deferredValidation": "257289"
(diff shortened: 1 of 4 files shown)


howydev (approved):
LGTM, but would recommend waiting for another approval

generally, since sig val can be used to execute arbitrary actions on the account via deferred actions, isSignatureValidation is a very risky permission that should be controlled very tightly right?

Zer0dot (approved):

PR #279, opened 2024-11-07
fix: [QS-ALC-2] add non zero address check for paymaster in modules

fangting-alchemy:
Address ALC 2 where 0 paymaster address was not checked in modules.
* PaymasterGuardModule,  zero address paymaster is not allowed to be installed.
* NativeTokenLimitModule, zero address paymaster is not allowed `preUserOpValidationHook`.

### DIFF
--- a/src/modules/permissions/NativeTokenLimitModule.sol
+++ b/src/modules/permissions/NativeTokenLimitModule.sol
@@ -73,10 +73,12 @@ contract NativeTokenLimitModule is ModuleBase, IExecutionHookModule, IValidation
             if (paymaster == address(0)) {
                 revert InvalidPaymaster();
             } else if (specialPaymasters[paymaster][msg.sender]) {
-                _decreaseLimit(entityId, userOp);
+                // Special paymaster specified, decrease limit
+                _decreaseLimit(entityId, userOp, true);
             }
         } else {
-            _decreaseLimit(entityId, userOp);
+            // No paymaster specified, decrease limit
+            _decreaseLimit(entityId, userOp, false);
         }
 
         return 0;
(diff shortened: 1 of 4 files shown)


src/modules/permissions/NativeTokenLimitModule.sol
```diff
@@ -163,4 +153,23 @@ contract NativeTokenLimitModule is ModuleBase, IExecutionHookModule, IValidation
     function supportsInterface(bytes4 interfaceId) public view override(ModuleBase, IERC165) returns (bool) {
         return interfaceId == type(IExecutionHookModule).interfaceId || super.supportsInterface(interfaceId);
     }
+
+    function _decreaseLimit(uint32 entityId, PackedUserOperation calldata userOp) internal {
+        uint256 vgl = UserOperationLib.unpackVerificationGasLimit(userOp);
+        uint256 cgl = UserOperationLib.unpackCallGasLimit(userOp);
+        uint256 pvgl;
+        uint256 ppogl;
+        if (userOp.paymasterAndData.length > 0) {
```
howydev: really small nit: since we're conditionally calling _decreaseLimit with the userOp.paymasterAndData.length condition check already, maybe we can refactor to:
1. pass in a bool instead
2. perform unpackPaymasterStaticFields in the top function and pass in `pvgl` + `ppogl` in the nonzero length case, or pass in `0` in the zero case

howydev (approved):
thanks for creating this, it helped me better understand the issue. have 1 nit/optional suggestion but otherwise fix looks good

0xrubes:
Putting it here for visibility too:
We would suggest another tweak for ALC-2 in the `PaymasterGuardModule`. Due to the fact that the `onInstall()` callback at installation is optional, the `paymasters[entityId][msg.sender]` mapping could remain unset at installation, therefore potentially still resolving to the zero slot. This means that the `paymasterAndData` field could still encode a zero address, opening up the attack vector of the module draining the account funds. While this would be somewhat of a configuration issue (as the `onInstall()` callback would need to be skipped), we would suggest also checking that `payingPaymaster != address(0x0)` in the `preUserOpValidationHook()` function.

PR #280, opened 2024-11-08
fix: [CL-009] address allowlist can be set independently of spend limits

howydev:


src/modules/permissions/AllowlistModule.sol
```diff
@@ -203,7 +203,9 @@ contract AllowlistModule is IExecutionHookModule, IValidationHookModule, ModuleB
                     setSelectorAllowlist(entityId, address(0), input.selectors[j], true);
                 }
             } else {
-                setAddressAllowlist(entityId, input.target, true, input.hasSelectorAllowlist);
+                setAddressAllowlist(
+                    entityId, input.target, true, input.hasSelectorAllowlist, input.hasERC20SpendLimit
+                );
                 updateLimits(entityId, input.target, input.hasERC20SpendLimit, input.erc20SpendLimit);
```
adamegyed, line 209: Will this line do a duplicate write?

howydev: not compared to the original, `hasSelectorAllowlist` and `hasERC20SpendLimit` are bools packed into the same storage slot

adamegyed (approved):
Looks good!

PR #281, opened 2024-11-08
fix: [CL-008] remove duplicate wrapping

howydev:
to reviewers: best to review other deferred action changes first since that can impact this


src/account/SemiModularAccountBase.sol
```diff
@@ -112,7 +112,13 @@ abstract contract SemiModularAccountBase is ModularAccountBase {
         if (sigValidation.eq(FALLBACK_VALIDATION)) {
             address fallbackSigner = _getFallbackSigner();
 
-            if (_checkSignature(fallbackSigner, _replaySafeHash(hash), signature)) {
+            // If called during validateUserOp, this implies that we're doing a deferred validation installation.
+            // In this case, as the hash is already replay-safe, we don't need to wrap it.
+            if (msg.sig != this.validateUserOp.selector) {
```
adamegyed, line 117: I'm not sure this is the right way to do this switching logic, it seems ok on first glance but feels a bit fragile. Could we make the caller of `_exec1271Validation` perform the wrapping in the cases where it is needed, and skip it in the deferred action case?

Zer0dot: Could you elaborate on this point? 

Would you instead not wrap it in `validateUserOp()`'s deferred validation section, but wrap in `isValidSignature()`?

howydev: Those approaches require modifying  `_isValidSignature` or `_validateUserOp` in ModularAccountBase (function doesn't exist in SMA) which would be a huge change

Agree that it feels jank, but this is the least complex solution. Happy to change my mind if there's an easier solution to the above or if specific issues about this way of doing things are found

adamegyed: Good points, yes that makes sense. I'm ok with this solution.

adamegyed (approved):

PR #282, opened 2024-11-08
fix: [CL-007] [QS-13] add invalidateDeferredValidationInstallNonce to native selectors check

howydev:


src/helpers/NativeFunctionDelegate.sol
```diff
@@ -36,6 +36,7 @@ contract NativeFunctionDelegate {
         || selector == uint32(IModularAccountView.getExecutionData.selector)
             || selector == uint32(IModularAccountView.getValidationData.selector)
         // check against ModularAccount methods
-        || selector == uint32(ModularAccountBase.performCreate.selector);
+        || selector == uint32(ModularAccountBase.performCreate.selector)
+            || selector == uint32(ModularAccountBase.invalidateDeferredValidationInstallNonce.selector);
```
adamegyed: Do we also want to add `isValidSignature`, `executeUserOp`, and the receiver functions `(IERC721Receiver.onERC721Received.selector`, `IERC1155Receiver.onERC1155Received.selector`, `IERC1155Receiver.onERC1155BatchReceived.selector`) to this?

Zer0dot: Went ahead and did this, also refactored to no longer import `IModularAccount` to make one clean block of function selectors which are all from `ModularAccountBase`. Lmk what you think!

adamegyed (approved):
Looks good!

PR #283, opened 2024-11-08
fix: [QS-4] update bitmask to allow 6 bytes of info instead of 4

howydev:


howydev:
closing in favor of: https://github.com/alchemyplatform/modular-account/pull/285

PR #287, opened 2024-11-11
fix: [QS-01] revert on presence of validation hooks in deferred validation, and run validation-associated execution hooks

Zer0dot:
## Motivation

Currently, it's possible to skip userOp validation hooks if the validation function being used is also a signature validation through deferred installation. Furthermore, we're not running userOp validation-associated execution hooks, when we could be.

## Solution

Revert if the validation being used includes validation hooks, and run the validation-associated pre and post hooks tied to that validation.

### DIFF
--- a/src/account/ModularAccountBase.sol
+++ b/src/account/ModularAccountBase.sol
@@ -104,6 +104,7 @@ abstract contract ModularAccountBase is
     error UnexpectedAggregator(ModuleEntity validationFunction, address aggregator);
     error UnrecognizedFunction(bytes4 selector);
     error ValidationFunctionMissing(bytes4 selector);
+    error OuterValidationHasValidationHooks();
 
     // Wraps execution of a native function with runtime validation and hooks
     // Used for upgradeTo, upgradeToAndCall, execute, executeBatch, installExecution, uninstallExecution,
@@ -415,11 +416,18 @@ abstract contract ModularAccountBase is
         if (hasDeferredAction) {
             // Use outer validation to validate the UO, and the inner validation as 1271 validation for the
             // deferred action.
+            // Because this bypasses UO validation hooks, we require that the validation used does not include any
+            // validation hooks.
+            if (!getAccountStorage().validationStorage[validationFunction].validationHooks.isEmpty()) {
+                revert OuterValidationHasValidationHooks();
+            }
+
+            // Use outer validation as a 1271 validation, then use the inner validation to validate the UO.
 
             // Get the length of the deferred action data.
             uint256 encodedDataLength = uint32(bytes4(userOp.signature[25:29]));
 
-            // Load the pointer to the abi-encoded data.
+            // Load the pointer to the encoded data.
             bytes calldata encodedData = userOp.signature[29:29 + encodedDataLength];
 
             // Get the deferred action signature length.
@@ -443,8 +451,15 @@ abstract contract ModularAccountBase is
             // Update the validation data with the deadline.
             validationData = uint256(deadline) << 160;
 
+            // Run the validation associated execution hooks, this is extracted in a function for stack management.
+            DensePostHookData postHookData =
+                _runValidationAssociatedExecHooks(validationFunction, encodedData[63:]);
+
             // Perform the deferred action's self call on the account.
             ExecutionLib.callBubbleOnRevertTransient(address(this), 0, encodedData[63:]);
+
+            // Do the cached post hooks
+            ExecutionLib.doCachedPostHooks(postHookData);
         } else {
             userOpSignature = userOp.signature[25:];
         }
@@ -506,7 +521,7 @@ abstract contract ModularAccountBase is
         emit DeferredActionNonceInvalidated(nonce);
 
         // Compute the typed data hash to verify the signature over
-        bytes32 typedDataHash = _computeDeferredValidationInstallTypedDataHash(
+        bytes32 typedDataHash = _computeDeferredActionTypedDataHash(
             encodedData[63:], // The encoded call without the nonce, deadline, and validation function
             nonce,
             deadline,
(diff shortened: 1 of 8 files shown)


src/account/ModularAccountBase.sol
```diff
@@ -413,12 +414,18 @@ abstract contract ModularAccountBase is
         ///      [(33 + deferredActionSigLength + encodedDataLength):] : bytes, userOpSignature. This is the
         ///         signature passed to the inner validation.
         if (hasDeferredAction) {
+            // Because this bypasses UO validation hooks, we require that the validation used does not include any
+            // validation hooks.
+            if (!getAccountStorage().validationStorage[validationFunction].validationHooks.isEmpty()) {
```
fangting-alchemy: We should document this during validation installation.

> If you want to use a validation to approve deferred actions in the future, it should not have validation hooks.

Zer0dot: Where do you think this fits? As a comment in `installValidation()`? Or perhaps in the documentation for integrators/SDK?

fangting-alchemy: In the documentation notion doc for now to later transfer to the doc site is enough probably.

Zer0dot: Added a section in the documentation, feel free to give it a peek!

src/account/ModularAccountBase.sol
```diff
@@ -439,12 +439,16 @@ abstract contract ModularAccountBase is
...
+            (uint48 deadline, DensePostHookData postHookData) =
+                _validateDeferredActionAndSetNonce(uoValidation, encodedData, deferredActionSig);
             // Update the validation data with the deadline.
             validationData = uint256(deadline) << 160;
 
             // Perform the deferred action's self call on the account.
             ExecutionLib.callBubbleOnRevertTransient(address(this), 0, encodedData[63:]);
+
+            // Do the cached post hooks
+            ExecutionLib.doCachedPostHooks(postHookData);
```
adamegyed: Does it make sense to move this section into the `_validateDeferredActionAndSetNonce` function? It seems like the functionality is leaking out from the function, as it runs the pre-exec hooks there. Or otherwise reorganize to do the pre-exec hooks + self-call + post exec hooks together?

(Side note: this is a case where we can manually clear the memory after running with `freezeFMP` / `restoreFMP`, because this action isn't reused for the user op validation later.)

src/account/ModularAccountBase.sol
```diff
@@ -479,48 +483,44 @@ abstract contract ModularAccountBase is
...
-            deadline,
-            userOpValidationFunction
-        );
+        // Compute the typed data hash to verify the signature over, and fetch the deadline
+        (bytes32 typedDataHash, uint48 deadline) = _checkAndCompute712Data(encodedData, userOpValidationFunction);
 
         // Clear the memory after performing signature validation
         MemSnapshot memSnapshot = MemManagementLib.freezeFMP();
-        if (_isValidSignature(defActionSigValidation.moduleEntity(), typedDataHash, sig) != _1271_MAGIC_VALUE) {
+        if (_isValidSignature(defActionValidationModuleEntity, typedDataHash, sig) != _1271_MAGIC_VALUE) {
```
adamegyed: Given that we assert there are no pre-validation hooks at all, we can also assert there are no pre-signature validation hooks, which means we will never have any per-hook data. Given that, I *think* we should be good to replace this call to `_isValidSignature` with `_exec1271Validation`, and omit the signature segment decoding. Wdyt?

src/account/ModularAccountBase.sol
```diff
@@ -711,6 +711,21 @@ abstract contract ModularAccountBase is
         return ExecutionLib.invokeUserOpCallBuffer(callBuffer, userOpValidationFunction, signatureSegment);
     }
 
+    function _runValidationAssociatedExecHooks(ModuleEntity validationFunction, bytes calldata callData)
```
adamegyed: Oh, this may be a good place to wrap all of {pre exec hooks, the transient self-call, and the post-exec hooks} while wrapping with the memory deallocator.

src/account/ModularAccountBase.sol
```diff
@@ -413,13 +414,12 @@ abstract contract ModularAccountBase is
         ///      [(33 + deferredActionSigLength + encodedDataLength):] : bytes, userOpSignature. This is the
         ///         signature passed to the inner validation.
         if (hasDeferredAction) {
-            // Use outer validation to validate the UO, and the inner validation as 1271 validation for the
-            // deferred action.
+            // Use outer validation as a 1271 validation, then use the inner validation to validate the UO.
```
adamegyed: This comment shouldn't be switched, right?

src/account/ModularAccountBase.sol
```diff
@@ -719,6 +734,74 @@ abstract contract ModularAccountBase is
...
+        // uint48 deadline,
+        bytes25 validationFunction
+    ) internal returns (bytes32, uint48) {
+        uint256 nonce = uint256(bytes32(encodedData[:32]));
+        uint48 deadline = uint48(bytes6(encodedData[32:38]));
+
+        // Check that the passed nonce isn't already invalidated.
+        if (getAccountStorage().deferredActionNonceUsed[nonce]) {
+            revert DeferredActionNonceInvalid();
+        }
```
adamegyed, line 744: This is correct to do in the workflow, but it isn't described by the function name - can we refactor this out to the parent function, or update the function naming?

adamegyed (commented):
Looks good overall, have a few questions about code structure

src/account/ModularAccountBase.sol
Zer0dot: Sounds good to me!

src/account/ModularAccountBase.sol
Zer0dot: Function name changed, it now handles the entire signature flow

src/account/ModularAccountBase.sol
```diff
@@ -475,50 +482,53 @@ abstract contract ModularAccountBase is
...
-        uint48 deadline = uint48(bytes6(encodedData[32:38]));
 
         ValidationConfig defActionSigValidation = ValidationConfig.wrap(bytes25(encodedData[38:63]));
         bool isGlobalSigValidation = defActionSigValidation.isGlobal();
 
+        ModuleEntity defActionValidationModuleEntity = defActionSigValidation.moduleEntity();
+
+        // Because this bypasses UO validation hooks, we require that the validation used does not include any
+        // validation hooks.
+        if (!getAccountStorage().validationStorage[defActionValidationModuleEntity].validationHooks.isEmpty()) {
```
adamegyed: We should use `validationStorage[defActionValidationModuleEntity].validationHookCount != 0`, to prevent the cold `sload` here.

src/account/ModularAccountBase.sol
Zer0dot: Should be good to go now, it's all in a `_handleDeferredAction()` function!

src/account/ModularAccountBase.sol
Zer0dot: I went ahead and wrapped the entire call to `_handleDeferredAction()` which is a new function that basically does everything except the initial decoding in `_validateUserOp()`. It looks good to me, but let me know if there's any memory-related issue you spot.

adamegyed (approved):
Looks good, 1 small storage optimization

src/account/ModularAccountBase.sol
Zer0dot: On it!

PR #288, opened 2024-11-11
feat: add 7702 compatible semi modular account

fangting-alchemy:
## Motivation

Support EIP-7702 EOA delegation with semi modular account.

## Solution
Default fallback signer to `address(this)` aka EOA address for 7702 EOA account.

### DIFF
--- a/src/account/SemiModularAccount7702.sol
+++ b/src/account/SemiModularAccount7702.sol
@@ -1,0 +1,37 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {IModularAccount} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+import {SemiModularAccountBase} from "./SemiModularAccountBase.sol";
+
+/// @title Semi-Modular Account for EIP-7702 EOAs
+/// @author Alchemy
+/// @notice An implementation of a semi-modular account which reads the signer as the address(this).
+/// @dev Inherits SemiModularAccountBase. This account can be used as the delegate contract of an EOA with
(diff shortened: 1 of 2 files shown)


Zer0dot (approved):

src/account/SemiModularAccount7702.sol
```diff
@@ -0,0 +1,37 @@
...
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+import {SemiModularAccountBase} from "./SemiModularAccountBase.sol";
+
+/// @title Semi-Modular Account for EIP-7702 EOAs
+/// @author Alchemy
+/// @notice An implementation of a semi-modular account which reads the signer as the address(this).
+/// @dev Inherits SemiModularAccountBase. This account can be used as the delegate contract of an EOA with
+/// EIP-7702, where address(this) (aka the EOA address) is the default fallback signer.
+contract SemiModularAccount7702 is SemiModularAccountBase {
```
jaypaik, line 14: [Discussed offline, but documenting here for alignment]

Only callout here is that ModularAccountBase has the `upgradeToAndCall` function which doesn't make sense for the 7702 flow. If we want to tidy things up a bit and reduce an external function, we can probably pull this function out into another base class and exclude it here. Is this worth it? cc @alchemyplatform/contractoors .

Zer0dot: Right, since we're not using any proxy layer, this makes sense. I wonder if it's worth the added "extra hops"

fangting-alchemy: It would just be a tiny bit code saved on 7702 SMAs. Require duplicated impl on both MA and other SMAs.
We could disable the upgrade for SMA7702 though. Updated.

Zer0dot: Nice!

PR #289, opened 2024-11-12
fix: [CL ALCHEMY-002] use correct validation function for running hooks in executeUserOp

adamegyed:
## Motivation

Addresses CL ALCHEMY-002.

When a deferred action is used, the validation-associated exec hooks, which run within `executeUserOp`, load the wrong set of hooks by reading from the first 24 bytes of `userOp.signature`, which refers to the validation of the deferred action, rather than the user op.

## Solution

When using deferred actions, switch the encoding locations of the validation for the user op and the validation for the deferred action. This results in `executeUserOp` loading and running the correct set of hooks.

Note that while the locations are switched, the contents of the EIP-712 struct for deferred actions remains the same, so the "outer" user op validation remains what is signed over.

Also add a test to verify this behavior.

### DIFF
--- a/src/account/ModularAccountBase.sol
+++ b/src/account/ModularAccountBase.sol
@@ -438,7 +438,7 @@ abstract contract ModularAccountBase is
             // The struct must sign over the user op validation function, nonce, deadline, and the deferred action.
             // Note that while the declared type of the UO validation is `ValidationConfig`, the flags are
             // interpretted as validation selection flags, not validation installation flags.
-            ValidationConfig uoValidation = ValidationConfig.wrap(bytes25(userOp.signature[:25]));
+            bytes25 uoValidation = bytes25(userOp.signature[:25]);
             uint48 deadline = _validateDeferredActionAndSetNonce(uoValidation, encodedData, deferredActionSig);
             // Update the validation data with the deadline.
             validationData = uint256(deadline) << 160;
@@ -476,7 +476,7 @@ abstract contract ModularAccountBase is
 
     /// @return The deadline of the deferred action
     function _validateDeferredActionAndSetNonce(
-        ValidationConfig userOpValidationFunction,
+        bytes25 userOpValidationFunction,
         bytes calldata encodedData,
         bytes calldata sig
     ) internal returns (uint48) {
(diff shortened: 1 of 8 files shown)


src/account/ModularAccountBase.sol
```diff
@@ -434,18 +435,16 @@ abstract contract ModularAccountBase is
                 userOp.signature[33 + encodedDataLength:33 + encodedDataLength + deferredActionSigLength];
 
             //Validate the signature.
-            (uint48 deadline, ValidationConfig newValidationFunction) = _validateDeferredActionAndSetNonce(
-                validationFunction, isGlobalValidation, encodedData, deferredActionSig
-            );
+            // The struct must sign over the user op validation function, nonce, deadline, and the deferred action.
+            // Note that while the declared type of the UO validation is `ValidationConfig`, the flags are
+            // interpretted as validation selection flags, not validation installation flags.
+            ValidationConfig uoValidation = ValidationConfig.wrap(bytes25(userOp.signature[:25]));
```
Zer0dot: Since it's not technically a ValidationConfig, would it make sense just to use a raw bytes25?

src/account/ModularAccountBase.sol
```diff
@@ -475,26 +474,26 @@ abstract contract ModularAccountBase is
...
+    ) internal returns (uint48) {
         // Decode stack vars for the deadline and nonce.
         // The deadline, nonce, inner validation, and deferred call selector are all at fixed positions in the
         // encodedData.
         uint256 nonce = uint256(bytes32(encodedData[:32]));
         uint48 deadline = uint48(bytes6(encodedData[32:38]));
 
-        ValidationConfig uoValidation = ValidationConfig.wrap(bytes25(encodedData[38:63]));
+        ValidationConfig defActionSigValidation = ValidationConfig.wrap(bytes25(encodedData[38:63]));
+        bool isGlobalSigValidation = defActionSigValidation.isGlobal();
```
Zer0dot, line 490: @adamegyed is this what you were referring to previously-- in that we treat this like a ValidationConfig when in reality it should be a ModuleEntity + validationFlags (where 0b00000001 is isGlobal)?

But in reality the flags are different.

adamegyed: Yes, exactly.

Zer0dot: Wdyt if we made this a bytes25 instead too?

Zer0dot (approved):

src/account/ModularAccountBase.sol
adamegyed: Update to `bytes25` here and in ModuleSignatureUtils!

PR #290, opened 2024-11-15
fix: [Quantstamp-ALC-14] fix execution data view for native functions

fangting-alchemy:
## Motivation
Execution data view method is not returning the correct fields for native functions. 

## Solution

Return correct fields for the following two types of native functions. Needed to create an interface to avoid circular dependency. It also helps the MA readability now that all external funcs have interfaces.
* Native view functions
* Other native functions

Also refactored `NativeFunctionDelegate` to import interfaces when available.

### DIFF
--- a/src/account/ModularAccountBase.sol
+++ b/src/account/ModularAccountBase.sol
@@ -27,6 +27,8 @@ import {UUPSUpgradeable} from "solady/utils/UUPSUpgradeable.sol";
 
 import {ExecutionInstallDelegate} from "../helpers/ExecutionInstallDelegate.sol";
 import {_coalescePreValidation, _coalesceValidation} from "../helpers/ValidationResHelpers.sol";
+
+import {IModularAccountBase} from "../interfaces/IModularAccountBase.sol";
 import {
     DensePostHookData,
     ExecutionLib,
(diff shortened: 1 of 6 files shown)


src/account/ModularAccountBase.sol
```diff
@@ -1015,6 +1002,11 @@ abstract contract ModularAccountBase is
         return _globalValidationAllowed(selector) && _isValidationGlobal(validationFunction);
     }
 
+    function _globalValidationAllowed(bytes4 selector) internal view virtual returns (bool) {
```
Zer0dot, line 1105: Small q-- why not combine this function into the one it calls in the ModularAccountView?

Zer0dot: Ohh looks like it's because we need a distinction between native & exec functions in the returned data from the view function, could we instead have this check for the global variable, then have the native function check for the native function variable? Let me write up what I mean on a comment in the ModularAccountView.

Zer0dot: Nvm, I was writing it and realized the point is actually to distinguish between native view functions and native non-view functions, lgtm!

Zer0dot (approved):

PR #291, opened 2024-11-15
fix: [QUANTSTAMP-ALC-20] Add NativeTokenLimitModule missing supported interface

Zer0dot:
## Motivation

We're missing a supported interface in `supportsInterface()`.

## Solution

Add the `IValidationHookModule` interface ID to the supported interface check.


adamegyed (approved):
Looks good!

PR #293, opened 2024-11-17
fix: [CL ALCHEMY-003] validation function uniqueness in nonce

adamegyed:
## Motivation

Addresses CL ALCHEMY-002.

Because the account does not restrict `userOp.nonce` to be unique per validation function, it is possible for two different validation function installations (i.e. two different keys) to have valid signature over different user ops with the same nonce. On networks without protected transactions, this poses a frontrunning risk in the event that an installed validation function is malicious.

To address this, some part of the nonce key must be unique to the validation being used. However, it should still be possible for a given validation function to have some number of parallel nonces at a time. This precludes setting the entire 24 bytes of the parallel nonce key to the validation function, because it would leave no space for the actual parallel nonce key.

Given that the nonce must be unique over the validation function being used, it no longer makes sense to encode the validation to use in the signature field, as it will always be a duplicate of the content in the nonce.

## Solution

Make validation entity ids unique over the entire account, except for the special casing with direct call validation. To support this structure, introduce a new data type called `ValidationLocator`, which represents a tagged union type between a unique validation entity id, and the address of a direct call validation's module.

Introduce encoding methods for putting the validation locator into either the signature of ERC-1271 validation, authorization of runtime validation, or the nonce for user op validation.

Update the storage references used by ModularAccountBase, SemiModularAccountBase, and the account's install and view functions.

Update the DeferredAction EIP-712 struct for the new switching format. This change is expected to be overwritten by the fix to CL ALCHEMY-004.

Exhaustively go through all test cases and update them to have unique validation entity ids, and to use the new format for specifying the validation.

### Reviewer recommendation

This PR is split up into multiple commits, due to the large scope of the changes. I recommend reviewing commits individually at first, to understand the steps taken.


src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,311 @@
...
+                if validationType {
+                    // Mask it to the lower 21 bytes
+                    result := and(nonce, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)
+                    break
+                }
+                // If not using direct call validation, the validation locator contains a 32-byte entity ID
+                // Mask it to the lower 5 bytes
+                result := and(nonce, 0xFFFFFFFFFF)
+                break
+            }
```
fangting-alchemy:
```
        switch and(nonce, IS_DIRECT_CALL_VALIDATION)
        case 1 { result := and(nonce, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF) }
        default { result := and(nonce, 0xFFFFFFFFFF)}
```

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,311 @@
...
+        assembly ("memory-safe") {
+            // Regular validation requires at least 5 bytes. Direct call validation requires at least 21 bytes,
+            // checked later.
+            if lt(signature.length, 5) { revert(0, 0) }
+
+            result := calldataload(signature.offset)
+
+            let validationOptions := shr(248, result)
+
+            // TODO: measure gas of switch vs for loop
```
fangting-alchemy: remove todo?

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,311 @@
...
+            switch and(validationOptions, IS_DIRECT_CALL_VALIDATION)
+            case 0 {
+                // If not using direct call validation, the validation locator contains a 32-byte entity ID
```
fangting-alchemy, line 117: should be 4 bytes entity ID

fangting-alchemy: (not sure why the propose change option is gone for me)

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,311 @@
...
+        result = uint256(uint160(directCallValidation)) << 8 | IS_DIRECT_CALL_VALIDATION;
+        if (isGlobal) {
+            result |= VALIDATION_TYPE_GLOBAL;
+        }
+        if (hasDeferredAction) {
+            result |= HAS_DEFERRED_ACTION;
+        }
+        // Finally, shift left to make space for the sequential nonce key
+        result <<= 64;
+    }
```
fangting-alchemy: Can remove `packNonce` and keep `packNonceDirectCall` for packing nonce.

src/helpers/Constants.sol
```diff
@@ -3,5 +3,10 @@ pragma solidity ^0.8.26;
 
 import {ModuleEntity} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
 
+import {ValidationLookupKey} from "../libraries/ValidationLocatorLib.sol";
+
 // Magic value for the ModuleEntity of the fallback validation for SemiModularAccount.
 ModuleEntity constant FALLBACK_VALIDATION = ModuleEntity.wrap(bytes24(0));
+
+// Magic value for the ValidationLookupKey of the fallback validation for SemiModularAccount.
+ValidationLookupKey constant FALLBACK_VALIDATION_LOOKUP = ValidationLookupKey.wrap(uint168(0));
```
fangting-alchemy:
```suggestion
ValidationLookupKey constant FALLBACK_VALIDATION_LOOKUP_KEY = ValidationLookupKey.wrap(uint168(0));
```

fangting-alchemy (approved):
Generally looks good. Some minor suggestions and typo/error in comments. 
Approving to speed things up.

src/account/AccountStorage.sol
```diff
@@ -1,8 +1,9 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.26;
 
-import {HookConfig, ModuleEntity} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
+import {HookConfig} from "@erc6900/reference-implementation/interfaces/IModularAccount.sol";
 
+import {ValidationLookupKey} from "../../src/libraries/ValidationLocatorLib.sol";
```
jaypaik:
```suggestion
import {ValidationLookupKey} from "../libraries/ValidationLocatorLib.sol";
```

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,327 @@
...
+
+    function hasDeferredAction(ValidationLocator locator) internal pure returns (bool) {
+        return (ValidationLocator.unwrap(locator) & _HAS_DEFERRED_ACTION) != 0;
+    }
+
+    function isDirectCallValidation(ValidationLocator locator) internal pure returns (bool) {
+        return (ValidationLocator.unwrap(locator) & _IS_DIRECT_CALL_VALIDATION) != 0;
+    }
+
+    function configToLookup(ValidationConfig validationConfig)
```
jaypaik:
```suggestion
    function configToLookupKey(ValidationConfig validationConfig)
```

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,327 @@
...
+        if (validationConfig.entityId() == DIRECT_CALL_VALIDATION_ENTITYID) {
+            result = ValidationLookupKey.wrap(
+                uint168(uint160(validationConfig.module())) << 8 | _IS_DIRECT_CALL_VALIDATION
+            );
+        } else {
+            result = ValidationLookupKey.wrap(uint168(uint160(validationConfig.entityId())) << 8);
+        }
+    }
+
+    function moduleEntityToLookup(ModuleEntity _moduleEntity) internal pure returns (ValidationLookupKey result) {
```
jaypaik:
```suggestion
    function moduleEntityToLookupKey(ModuleEntity _moduleEntity) internal pure returns (ValidationLookupKey result) {
```

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,327 @@
...
+        (address module, uint32 _entityId) = ModuleEntityLib.unpack(_moduleEntity);
+        if (_entityId == DIRECT_CALL_VALIDATION_ENTITYID) {
+            result = ValidationLookupKey.wrap(uint168(uint160(module)) << 8 | _IS_DIRECT_CALL_VALIDATION);
+        } else {
+            result = ValidationLookupKey.wrap(uint168(uint160(_entityId)) << 8);
+        }
+    }
+
+    function directCallLookup(address directCallValidation) internal pure returns (ValidationLookupKey result) {
```
jaypaik:
```suggestion
    function directCallLookupKey(address directCallValidation) internal pure returns (ValidationLookupKey result) {
```

src/libraries/ValidationLocatorLib.sol
adamegyed: Good call. I changed this to use a switch case in the later PR and didn't rebase it back into here, so it's already doing this at the top of the stack of commits.

src/libraries/ValidationLocatorLib.sol
adamegyed: removed at the top of the commit stack - I realize this may be an issue with reviewing individual commits, will make a note of this for later.

src/libraries/ValidationLocatorLib.sol
adamegyed: I'm not sure what this means - the two are used in different circumstances to pack the different types of validation locators.

I'll merge this PR now, but feel free to follow up on this.

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,335 @@
...
+    function directCallAddress(ValidationLocator locator) internal pure returns (address result) {
+        assembly ("memory-safe") {
+            result := and(shr(8, locator), 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF)
+        }
+    }
+
+    // Only safe to call if the locator has been asserted to be a non-direct call validation.
+    function entityId(ValidationLocator locator) internal pure returns (uint32 result) {
+        assembly ("memory-safe") {
+            result := and(shr(8, locator), 0xFFFFFFFFFFFFFFFF)
```
0xrubes, line 170: Why are we masking 8 bytes here? 4 should be sufficient. Implicit `uint32` cast does result in proper bits to be taken, so no real issue

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,335 @@
...
+        uint32 validationEntityId,
+        bool _isGlobal,
+        bool _hasDeferredAction,
+        bytes memory signature
+    ) internal pure returns (bytes memory result) {
+        uint8 options = 0;
+        if (_isGlobal) {
+            options |= _VALIDATION_TYPE_GLOBAL;
+        }
+        if (_hasDeferredAction) {
```
0xrubes, line 303: Aren't deferred actions exclusive to the UserOperation flow? Why would we need to set this option here for the 1271 signature validation and runtime validation flow?

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,335 @@
...
+    {
+        result = ValidationLookupKey.wrap(uint168(uint160(directCallValidation)) << 8 | _IS_DIRECT_CALL_VALIDATION);
+    }
+
+    function lookupKey(ValidationLocator locator) internal pure returns (ValidationLookupKey result) {
+        assembly ("memory-safe") {
+            result := and(locator, 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF04)
+        }
+    }
+
```
0xrubes, line 226: Why are we keeping the direct call validation flag here? Isn't that one implicit with the designated `DIRECT_CALL_VALIDATION_ENTITYID`?

0xrubes: Ah I see, I think it is due to the fact that the lookup key is only address + options flag instead of entityId + address to keep some unused bytes for the parallel nonce

src/libraries/ValidationLocatorLib.sol
```diff
@@ -0,0 +1,335 @@
...
+    // Operators
+
+    function eq(ValidationLookupKey a, ValidationLookupKey b) internal pure returns (bool) {
+        return ValidationLookupKey.unwrap(a) == ValidationLookupKey.unwrap(b);
+    }
+}
+
+using ValidationLocatorLib for ValidationLocator global;
```
nikita-quantstamp, line 334: Small nit pick, I think its better to move these two `using` to the top of the file to ease readability

src/libraries/ValidationLocatorLib.sol
howydev: doesn't the cast do the mask? do we need to mask it at all?

howydev: same with `directCallAddress`

PR #298, opened 2024-11-22
fix: [Quantstamp - ALC12 & S1] fix validUntil 0 value handing in timeRangeModule

fangting-alchemy:
…RangeModule

## Motivation

`validUntil == 0` was mishandled in `TimeRangeModule`. It should be allowed.

## Solution

Treat `validUntil` as `type(uint48).max` when value is 0.


src/modules/permissions/TimeRangeModule.sol
```diff
@@ -112,9 +113,14 @@ contract TimeRangeModule is IValidationHookModule, ModuleBase {
     /// @param validUntil The timestamp until which the time range is valid, inclusive.
     /// @param validAfter The timestamp after which the time range is valid, inclusive.
     function setTimeRange(uint32 entityId, uint48 validUntil, uint48 validAfter) public {
-        if (validUntil <= validAfter) {
+        if (validUntil == 0) {
+            if (validAfter == validUntil) {
+                revert TimeRangeNotValid();
+            }
```
adamegyed, line 119: Why do we disallow this case?

adamegyed (approved):
Looks good, not sure if we need 1 of the inner checks

src/modules/permissions/TimeRangeModule.sol
jaypaik: Maybe because of this: https://github.com/eth-infinitism/account-abstraction/blob/6f02f5a28a20e804d0410b4b5b570dd4b076dcf9/contracts/core/EntryPoint.sol#L600-L602

ChainLight called this out in their comment.

fangting-alchemy: In this case `validAfter == validUntil` means they are both 0.
Meaning, everything passes. In that case, this hooks is useless and should not be set.

PR #299, opened 2024-11-22
fix: [CL ALCHEMY-004] enforce unique deferred action nonce per UO validation

adamegyed:
## Motivation

Addresses CL ALCHEMY-004.

Deferred action nonces are not unique per user op validation, which means a malicious validation may front-run deferred actions with duplicates using the same nonce, causing a potential DOS if a key is leaked.

## Solution

Replace the custom nonce in the deferred action EIP-712 struct with `userOp.nonce`. This ensures that the deferred action is still signing over the user op validation, as that is also encoded in the nonce.

This also lets us remove the custom nonce management system, as invalidations can be done using the entrypoint's `NonceManager` features.


src/account/SemiModularAccountBase.sol
```diff
@@ -61,6 +63,18 @@ abstract contract SemiModularAccountBase is ModularAccountBase {
...
 
+    function installValidation(
+        ValidationConfig validationConfig,
+        bytes4[] calldata selectors,
+        bytes calldata installData,
+        bytes[] calldata hooks
+    ) external override wrapNativeFunction {
+        if (validationConfig.entityId() == FALLBACK_VALIDATION_ID && validationConfig.module() != address(0)) {
+            revert FallbackSignerMismatch();
+        }
```
jaypaik: Mind leaving a comment for context?

adamegyed: Previously, it was possible to "alias" the fallback validation by installing a module at the reserved validation entity id `0`. With the organizational changes to the account, not failing here could cause unexpected behavior, so this change was made to explicitly revert and warn the caller that this operation would not do what is requested.

Note that this state can still be reached by upgrading from MA to SMA, but should be handled with initialization / de-init steps.

src/account/SemiModularAccountBase.sol
```diff
@@ -61,6 +63,18 @@ abstract contract SemiModularAccountBase is ModularAccountBase {
...
     }
 
+    function installValidation(
+        ValidationConfig validationConfig,
+        bytes4[] calldata selectors,
+        bytes calldata installData,
+        bytes[] calldata hooks
+    ) external override wrapNativeFunction {
+        if (validationConfig.entityId() == FALLBACK_VALIDATION_ID && validationConfig.module() != address(0)) {
+            revert FallbackSignerMismatch();
```
fangting-alchemy: Optional:
```suggestion
            revert FallbackValidationInstallationNotAllowed();
```

fangting-alchemy (approved):

src/account/ModularAccountBase.sol
```diff
@@ -83,7 +81,7 @@ abstract contract ModularAccountBase is
     bytes32 internal constant _DOMAIN_SEPARATOR_TYPEHASH =
         0x47e79534a245952e8b16893a336b85a3d9ea9fa8c573f3d803afb92a79469218;
 
-    // keccak256("DeferredAction(uint256 nonce,uint48 deadline,uint168 validationLocator,bytes call)")
+    // keccak256("DeferredAction(uint256 nonce,uint48 deadline,bytes call)")
     bytes32 internal constant _DEFERRED_ACTION_TYPEHASH =
         0xa0a7682d80b18373e0ec8549850bc960f46562e16ae4ddee86ab9a3cc3d3a07c;
```
jaypaik:
```suggestion
        0x9b23e06584efc6b65fc854cee55011d89f86485487b6db36aed7d23884711ea3;
```

jaypaik (commented):
Generally would be nice to have consistency of calling things `lookupKey` vs `lookup`.

src/account/ModularAccountBase.sol
adamegyed: Good eye, this should also be corrected in the test file that computes it from the string.

src/account/SemiModularAccountBase.sol
adamegyed: Oh, I realize now this was to leave a code comment - will do

adamegyed:
Updated all occurrences of "lookup" to "lookupKey"

src/account/ModularAccountBase.sol
```diff
@@ -344,18 +339,12 @@ abstract contract ModularAccountBase is
...
-        getAccountStorage().deferredActionNonceUsed[nonce] = true;
-        emit DeferredActionNonceInvalidated(nonce);
-    }
-
     /// @inheritdoc IERC1271
     function isValidSignature(bytes32 hash, bytes calldata signature) external view override returns (bytes4) {
         (ValidationLocator locator, bytes calldata signatureRemainder) =
             ValidationLocatorLib.loadFromSignature(signature);
 
-        return _isValidSignature(locator, hash, signatureRemainder);
```
0xrubes, line 358: The options flag for validation modules in some cases are slightly unchecked. For example, I could request whether a certain signature of a non-global validation module is valid with the options flag set to it being a global validation. 
Perhaps, this could lead to wrong conclusions on the module/signature verification requesting side. Here the cases I observed:
* Signature validation can have the global validation flag incorrectly set without any reverts in the validation
* Signature validation and inner deferred action validation can have the `hasDeferredAction` flag set without consequences.
* Runtime validation can have the `hasDeferredAction` flag set without consequences.

I don't think these are very significant aspects, but still wanted to point them out 🙌

howydev: Thought about this a little, I think the only issue is being able to change a 0 byte to a nonzero byte so the UO consumes a little more gas in the case which the global validation flag is not set. I think we're comfortable leaving this in since any check would add more than 12 gas

PR #310, opened 2024-11-25
fix: [Quantstamp-ALC12] update state to max if setting timeRange until to 0 in module if necessary

fangting-alchemy:
## Solution

Store `validUntil` as `max` in TimeRangeModule when conditions are met:
* `validAfter` is not 0.
* `validUntil` is 0.

Revert defaulting to max when `validUntil == 0` in runtime hook.


jaypaik (approved):
lgtm pending `pnpm gas`

PR #316, opened 2024-12-06
feat: Add Module Deployment Address Prediction Script

Zer0dot:
## Motivation

Simple PR to add a script to print out the module addresses predicted using the salts from the environment (or zero by default).

This makes it easier to run deployment scripts by providing an easy to copy-paste solution for printing out all expected module addresses for the module deploy script.

## Solution

Add a script which prints out expected module addresses.

### DIFF
--- a/script/PredictModules.s.sol
+++ b/script/PredictModules.s.sol
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {console} from "forge-std/console.sol";
+
+import {Artifacts} from "./Artifacts.sol";
+import {ScriptBase} from "./ScriptBase.sol";
+import {Create2} from "@openzeppelin/contracts/utils/Create2.sol";
+
+// Predicts addresses for all standalone modules with salts taken from the environment.
+// - AllowlistModule
+// - NativeTokenLimitModule
+// - PaymasterGuardModule
+// - SingleSignerValidationModule
+// - TimeRangeModule
+// - WebAuthnValidationModule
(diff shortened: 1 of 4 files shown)


script/PredictModules.s.sol
```diff
@@ -0,0 +1,89 @@
...
+        singleSignerValidationModuleSalt = vm.envOr("SINGLE_SIGNER_VALIDATION_MODULE_SALT", uint256(0));
+        timeRangeModuleSalt = vm.envOr("TIME_RANGE_MODULE_SALT", uint256(0));
+        webAuthnValidationModuleSalt = vm.envOr("WEBAUTHN_VALIDATION_MODULE_SALT", uint256(0));
+    }
+
+    function run() public view onlyProfile("optimized-build") {
+        console.log("******** Logging Expected Module Addresses With Env Salts *********");
+
+        console.log(
+            "ALLOWLIST_MODULE=",
```
adamegyed, line 42: tiny lil nit, could we do `: ` instead of `=` for consistency?

adamegyed (approved):
Looks good. I'm not sure if this will be needed in the final deployment flow, as the output of create2crunch will tell us the expected addresses, and the deploy scripts will also warn if the predicted address has a mismatch, but it's useful to have on hand for quick testing.

script/PredictModules.s.sol
Zer0dot: Ah, the only reason I'm using : instead of = is to be easily pasted into an env file, wdyt?

adamegyed: oh I see. yeah that sounds good

PR #318, opened 2024-12-07
feat: Factory Deployment Script

Zer0dot:
## Motivation

We need scripts to predict the factory address and to deploy it. 

## Solution

Implement the aforementioned scripts, as well as some minor refactors to reduce resulting code duplication.

## TODO

- [x] Create the factory address prediction script.
- [x] Test the deployment

### DIFF
--- a/script/DeployFactory.s.sol
+++ b/script/DeployFactory.s.sol
@@ -41,8 +41,8 @@ contract DeployFactoryScript is ScriptBase, Artifacts {
         factoryOwner = _getFactoryOwner();
 
         // Load the expected address and salt from env vars.
-        expectedFactoryAddr = vm.envOr("FACTORY", address(0));
-        factorySalt = vm.envOr("FACTORY_SALT", uint256(0));
+        expectedFactoryAddr = vm.envOr("ACCOUNT_FACTORY", address(0));
+        factorySalt = vm.envOr("ACCOUNT_FACTORY_SALT", uint256(0));
     }
 
     function run() public onlyProfile("optimized-build") {
--- a/script/DeployFactory.s.sol
+++ b/script/DeployFactory.s.sol
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {console} from "forge-std/console.sol";
+
+import {ModularAccount} from "../src/account/ModularAccount.sol";
+import {SemiModularAccountBytecode} from "../src/account/SemiModularAccountBytecode.sol";
+import {Artifacts} from "./Artifacts.sol";
+import {ScriptBase} from "./ScriptBase.sol";
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+// Deploys the Account Factory. This requires the following env vars to be set:
+// - ENTRY_POINT
+// - MODULAR_ACCOUNT_IMPL
+// - SEMI_MODULAR_ACCOUNT_BYTECODE_IMPL
+// - SINGLE_SIGNER_VALIDATION_MODULE
+// - WEBAUTHN_VALIDATION_MODULE
+// - FACTORY_OWNER
+contract DeployFactoryScript is ScriptBase, Artifacts {
+    // State vars for expected addresses and salts.
+
+    address public expectedFactoryAddr;
+    uint256 public factorySalt;
+
+    // State vars for factory dependencies
+
+    IEntryPoint public entryPoint;
+    ModularAccount public modularAccountImpl;
+    SemiModularAccountBytecode public semiModularAccountBytecodeImpl;
+    address public singleSignerValidationModule;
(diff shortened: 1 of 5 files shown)


Zer0dot:
Note that the factory address prediction reads addresses from the env for dependencies, rather than using the predicted addresses in the same script, as those may differ from the predicted addresses.

script/DeployFactory.s.sol
```diff
@@ -0,0 +1,134 @@
...
+        entryPoint = _getEntryPoint();
+        modularAccountImpl = _getModularAccountImpl();
+        semiModularAccountBytecodeImpl = _getSemiModularAccountBytecodeImpl();
+        singleSignerValidationModule = _getSingleSignerValidationModule();
+        webAuthnValidationModule = _getWebAuthnValidationModule();
+        factoryOwner = _getFactoryOwner();
+
+        // Load the expected address and salt from env vars.
+        expectedFactoryAddr = vm.envOr("FACTORY", address(0));
+        factorySalt = vm.envOr("FACTORY_SALT", uint256(0));
```
adamegyed: I think the current var used in the `.env.example` file is `ACCOUNT_FACTORY` and `ACCOUNT_FACTORY_SALT`. I'm fine with using either that or the one here, but could you make it consistent across the locations?

script/GetInitcodeHash.s.sol
```diff
@@ -47,16 +47,8 @@ contract GetInitcodeHashScript is ScriptBase, Artifacts {
...
-        if (address(entryPoint) == address(0)) {
-            console.log(
-                "Env Variable 'ENTRYPOINT' not found or invalid, defaulting to v0.7 EntryPoint at "
-                "0x0000000071727De22E5E9d8BAf0edAc6f37da032"
-            );
-            entryPoint = IEntryPoint(0x0000000071727De22E5E9d8BAf0edAc6f37da032);
-        } else {
-            console.log("Using user-defined EntryPoint at: %x", address(entryPoint));
-        }
+        IEntryPoint entryPoint = _getEntryPoint(); //IEntryPoint(payable(vm.envOr("ENTRYPOINT", address(0))));
```
adamegyed: nit: delete comment

script/ScriptBase.sol
```diff
@@ -17,6 +20,41 @@ abstract contract ScriptBase is Script {
...
+
+        return entryPoint;
+    }
+
+    function _getModularAccountImpl() internal view returns (ModularAccount) {
+        return ModularAccount(payable(vm.envOr("MODULAR_ACCOUNT_IMPL", address(0))));
+    }
+
+    function _getSemiModularAccountBytecodeImpl() internal view returns (SemiModularAccountBytecode) {
+        return SemiModularAccountBytecode(payable(vm.envOr("SEMI_MODULAR_ACCOUNT_BYTECODE_IMPL", address(0))));
```
adamegyed, line 43: I think there's a different format already set up in `.env.example` without the `_IMPL` ending, could you make that consistent here too?

test/script/DeployFactory.s.t.sol
```diff
@@ -0,0 +1,86 @@
...
+    function setUp() public {
+        _deployFactoryScript = new DeployFactoryScript();
+
+        bytes32 zeroSalt = bytes32(0);
+
+        entryPoint = IEntryPoint(address(6));
+        modularAccountImpl = ModularAccount(payable(address(1)));
+        semiModularAccountBytecodeImpl = SemiModularAccountBytecode(payable(address(2)));
+        singleSignerValidationModule = address(3);
+        webAuthnValidationModule = address(4);
```
adamegyed: Why not deploy real copies of them? The EP can use the deploy script in OptimizedTest.

test/script/DeployFactory.s.t.sol
```diff
@@ -0,0 +1,86 @@
...
+        factoryOwner = address(5);
```
adamegyed: nit: prefer `makeAddr` over incrementing numbers due to appearing as precompiles in the test traces.

adamegyed (approved):
Looks good, had a couple small comments

test/script/DeployFactory.s.t.sol
Zer0dot: Honestly, could go either way, I just didn't really want to test anything related to deploying anything else-- but it's a small change, I'll address this!

script/ScriptBase.sol
Zer0dot: Yup, will do!

script/DeployFactory.s.sol
Zer0dot: Will do! Good catch, in my infinite wisdom I didn't check .env.example :)

script/ScriptBase.sol
Zer0dot: Oops-- actually the one in `.env.example` has `_IMPL`, I'd say we're good to keep it, it ensures the developer knows we're looking for an implementation, not an actual account.

PR #320, opened 2024-12-10
feat: Account Deployment Script

Zer0dot:
## Motivation

We need a reproducible way to predictably deploy our account implementations to any chain.

## Solution

Implement a script to deploy the execution install delegate and all account implementations.

### DIFF
--- a/script/DeployAccounts.s.sol
+++ b/script/DeployAccounts.s.sol
@@ -1,0 +1,121 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {console} from "forge-std/console.sol";
+
+import {ExecutionInstallDelegate} from "../src/helpers/ExecutionInstallDelegate.sol";
+import {Artifacts} from "./Artifacts.sol";
+import {ScriptBase} from "./ScriptBase.sol";
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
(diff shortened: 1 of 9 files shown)


script/DeployAccounts.s.sol
```diff
@@ -0,0 +1,121 @@
...
+import {ScriptBase} from "./ScriptBase.sol";
+import {IEntryPoint} from "@eth-infinitism/account-abstraction/interfaces/IEntryPoint.sol";
+
+// Deploys the three account implementations and an execution install delegate. This requires the following env
+// vars to be set:
+// - ENTRY_POINT (optional)
+contract DeployAccountsScript is ScriptBase, Artifacts {
+    // State vars for expected addresses and salts.
+
+    // uint256
```
adamegyed: remove

script/DeployAccounts.s.sol
```diff
@@ -0,0 +1,121 @@
...
+
+        // At this point, the delegate and entrypoint are valid, so we can safely proceed with
+        // using them as parameters and accessing them in wrapped functions.
+
+        _safeDeploy(
+            "Modular Account Impl",
+            expectedModularAccountImpl,
+            modularAccountImplSalt,
+            _getModularAccountInitcode(
+                IEntryPoint(entryPoint), ExecutionInstallDelegate(expectedExecutionInstallDelegate)
```
adamegyed: possibly redundant casting? It looks like `entryPoint` is already of type `IEntryPoint`.

This occurs here and below

adamegyed (approved):

PR #321, opened 2024-12-11
feat: deploy and address prediction scripts

adamegyed:
## Motivation

We need the ability to deploy and predict addresses of our deployments

## Solution

Add scripts to deploy all top-level contracts needed by our account implementation.

Add scripts to predict the deployment addresses of those contracts.

Add tests for each of these scripts.

### DIFF
--- a/script/ScriptBase.sol
+++ b/script/ScriptBase.sol
@@ -1,12 +1,22 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.26;
 
-import {ScriptBase} from "forge-std/Base.sol";
+import {Script} from "forge-std/Script.sol";
 import {console} from "forge-std/console.sol";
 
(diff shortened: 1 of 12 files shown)


script/ScriptBase.sol
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    function _safeDeploy(
+        string memory contractName,
+        address expectedAddress,
+        uint256 salt,
+        bytes memory creationCode,
+        function (bytes32) internal returns (address) deployFunction
+    ) internal {
+        console.log(string.concat("Deploying ", contractName, " with salt: ", vm.toString(salt)));
```
howydev, line 75: nit: from print statements I'm not able to tell if the script got the expected address or it's using 0, could we add the expected addr here too?

howydev (approved):
1 small nit otherwise lgtm

PR #324, opened 2024-12-13
feat: specialize deploy scripts per artifact

adamegyed:
## Motivation

- We want to set different compiler settings per deployed contract.
- CI formatting is broken since an upstream foundry update
- We were missing the 7702 account implementation from the deploy script

## Solution

- Add the Sma7702 deployment to the DeployAccounts script.
- Move the deployment of `ExecutionInstallDelegate` to the same script as the module deployments, and rename to `DeployStandalones` (i.e. not having dependencies).
   - Create a new foundry profile with higher optimizer runs for these contracts
- Move the deployment of SmaStorageOnly to a new deployment script and create its own profile with the same optimizer runs as the current setting.
- Increase the baseline optimizer runs for the profile `optimized-build`.
   -  Update the gas profile to use the same number of runs, as it only tests contracts using these settings.
- Adjust test cases to cover the new deploy script behavior
- Adjust the filtering on the sizes action in CI. This is not comprehensive anymore, since there need to be multiple profiles pulled, and there's still an outstanding task to fix the formatting, but it is a first step.

### Alternative considered

At first, I tried to make use of the new compilation restrictions feature in foundry, but it created duplicate compilation artifacts for the scripts, and there wasn't a mechanism to control which instance was used in the script. If this feature stabilizes, we should switch to it: https://github.com/foundry-rs/foundry/pull/8668


howydev (approved):
LGTM. curious - why are we using a separate script for sma storage?

adamegyed:
SMA-storage is ~500 bytes larger than the other variants, so I split it up to be able to bump up the optimizer runs on the other implementations

howydev (approved):

PR #353, opened 2025-12-18
feat: make replaySafeHash wrapping optional with validation option

dancoombs:


src/account/ModularAccountBase.sol
```diff
@@ -396,6 +400,20 @@ abstract contract ModularAccountBase is
...
 
+    /// @notice Returns the replay-safe hash generated from the passed typed data hash for 1271 validation.
+    /// @param hash The typed data hash to wrap in a replay-safe hash.
+    /// @return The replay-safe hash, to be used for 1271 signature generation.
+    ///
+    /// @dev Generates a replay-safe hash to wrap a standard typed data hash. This prevents replay attacks by
+    /// enforcing the domain separator, which includes this contract's address, the chainId, and the validation
+    /// module & entity id. This is only relevant for 1271 validation because UserOp validation relies on the UO
+    /// hash and the Entrypoint has safeguards.
+    function replaySafeHash(bytes32 hash, ModuleEntity validationModuleEntity) public view returns (bytes32) {
```
adamegyed: Making this public will increase bytecode size, and isn't strictly necessary (plus, doing an `eth_call` during txn prep is usually slow and requires implementation knowledge anyways.

But, this would make it easier to debug.

adamegyed (approved):
LGTM, had one callout / design decision note

